PR #21, opened 2026-06-09
Map connector username to DB user in ATTACH SQL

### LINKED ISSUE #20: Bug: Username is ignored during database attach
## What happened

Currently, DuckDB fails to attach databases such as MySQL, PostgreSQL, etc. when a username is provided.

After checking the plan, it seems that the provided username is not being mapped to the `user` parameter.

## Steps to reproduce

1. Create MySQL or Postgres connection with a username/password.
2. Open the Plan tab.
3. Check the generated `ATTACH` SQL.

## Environment

- Duckle version / release tag:
- OS + architecture (e.g. Windows 11 x64, macOS arm64): macOS arm64
- For a database connector: which DB and version (e.g. Oracle 21c XE, SQL Server 2022): mysql, postgres, etc.

kyounghoonJang:
## What this changes

This maps the username passed from the frontend to `user` during attach.

## Related issue

Closes #20 

## Type

- [x] Bug fix
- [ ] Performance
- [ ] Docs
- [ ] Refactor / internal

## Checklist

- [x] `cargo fmt` and `cargo clippy --workspace --all-targets -- -D warnings` pass
- [x] `npm --prefix frontend run lint` passes (if the frontend changed)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed PostgreSQL database connection handling to support multiple authentication credential naming conventions. The system now properly recognizes both credential parameter formats and correctly processes them when executing database commands, improving compatibility with different source configuration methods and ensuring consistent behavior across various credential specification formats.

### DIFF
--- a/crates/duckdb-engine/src/plan/builders.rs
+++ b/crates/duckdb-engine/src/plan/builders.rs
@@ -3160,7 +3160,10 @@ pub(crate) fn db_attach(props: &JsonValue, extension: &str, default_port: u64, r
     if let Some(db) = string_prop(props, "database").filter(|s| !s.is_empty()) {
         parts.push(format!("{}={}", db_key, db));
     }
-    if let Some(u) = string_prop(props, "user").filter(|s| !s.is_empty()) {
+    if let Some(u) = string_prop(props, "user")
+        .or_else(|| string_prop(props, "username"))
+        .filter(|s| !s.is_empty())
+    {
         parts.push(format!("user={}", u));
     }
     if let Some(p) = string_prop(props, "password").filter(|s| !s.is_empty()) {
--- a/crates/duckdb-engine/tests/execution.rs
+++ b/crates/duckdb-engine/tests/execution.rs
@@ -908,6 +908,39 @@ fn compiled_sql_redacts_secrets() {
 }
 
 #[test]
+fn compiled_sql_maps_username_to_attach_user() {
+    // The UI writes DB login names as `username`, while DuckDB's
+    // Postgres/MySQL connection string expects `user=...`.
+    // This is pure compilation; no live database is needed.
+    use duckle_duckdb_engine::compile_pipeline_sql_opts;
+    let d = doc(
+        json!([
+            node("s", "src.postgres", json!({
+                "host": "db.example.com",
+                "port": 5432,
+                "database": "app",
+                "username": "admin",
+                "password": "sup3rs3cr3tpw",
+                "tableName": "orders"
+            })),
+        ]),
+        json!([]),
+    );
+
+    let stages = compile_pipeline_sql_opts(&d, false).expect("compile_pipeline_sql_opts");
+    let all_sql = stages
+        .iter()
+        .map(|s| s.sql.clone())
+        .collect::<Vec<_>>()
+        .join("\n");
+    assert!(
+        all_sql.contains("user=admin"),
+        "expected username to be rendered as DB user in ATTACH SQL: {}",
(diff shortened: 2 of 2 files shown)


crates/duckdb-engine/src/plan/builders.rs
```diff
@@ -3160,7 +3160,7 @@ pub(crate) fn db_attach(props: &JsonValue, extension: &str, default_port: u64, r
     if let Some(db) = string_prop(props, "database").filter(|s| !s.is_empty()) {
         parts.push(format!("{}={}", db_key, db));
     }
-    if let Some(u) = string_prop(props, "user").filter(|s| !s.is_empty()) {
+    if let Some(u) = string_prop(props, "username").filter(|s| !s.is_empty()) {
```
SouravRoy-ETL:
```suggestion
    if let Some(u) = string_prop(props, "user").or_else(|| string_prop(props, "username")).filter(|s| !s.is_empty()) {
```
Read both keys: prefer `user` (the canonical key the rest of the engine and the integration tests use) and fall back to `username` (what the GUI writes). This fixes #20 while keeping `user`-keyed pipelines and tests working. Same pattern as `src.ftp` / `snk.ftp` in `plan/mod.rs`.

SouravRoy-ETL (commented):
Thanks for the fix, and for the clear repro in #20.

The diagnosis is right: the GUI writes the credential under the `username` prop key (the shared DB connection form via `credentialFields()` in `frontend/src/workflow-ui/fields/manifest-synth.ts`, the saved-connection editor, and `onPickConnection` all use `username`), while `db_attach()` read `user`. So for the Postgres / MySQL / MariaDB / CockroachDB / Redshift / pgvector connectors the username was silently dropped from the connection string. That is exactly issue #20, and `db_attach` is the only place these connectors read that property.

One change is needed before this can merge. The patch swaps the key outright (`user` to `username`) instead of accepting both, and `user` is still a live input on other paths:

- The engine integration tests for these exact connectors build props with the `user` key and assert success (`crates/duckdb-engine/tests/execution.rs`, e.g. the postgres / mysql / cockroach sink+source roundtrips). They are gated behind a live database (`DUCKLE_PG_HOST` / `DUCKLE_MYSQL_HOST`), so they pass in plain CI but fail against a real DB once `user` is no longer read.
- Any pipeline JSON authored with the historical `user` key would lose its login.
- The rest of the engine treats `user` as canonical. The only sites that accept `username` do so as a fallback (the File Transfer connectors in `plan/mod.rs`), never as a swap.

The safe fix is to read both, preferring `user` and falling back to `username`. That fixes #20 without regressing the `user` callers, and matches the existing fallback pattern. See the inline suggestion (you can apply it directly).

How to update and resubmit (no new PR needed, just push to this branch):

1. Apply the inline suggestion, or edit the line in `crates/duckdb-engine/src/plan/builders.rs` to the dual-read.
2. Optional but welcome: add a compile-only regression test that builds a `src.postgres` node with the `username` key and asserts the generated ATTACH string contains `user=...`, so this path stays covered. `compiled_sql_redacts_secrets` in `tests/execution.rs` is a good template (it needs no live DB).
3. Run `cargo fmt` and `cargo clippy --workspace --all-targets -- -D warnings`, commit (keep your sign-off), and push to the same branch. The PR updates automatically.

Unrelated heads-up: `src.email` (IMAP) has the same class of mismatch (the form writes `username`, the engine reads `user` and errors "user required"). Probably worth its own issue rather than folding it into this PR.

Issue #92, opened 2026-06-23
Feature request: GUI hot-reload from disk, MCP update_pipeline, and external AI API support


cola3k:
## Summary

Three related improvements to bridge the MCP ↔ GUI gap and make the built-in AI configurable.

## Background

Duckle ships with a local Qwen Coder 1.5B model that can generate pipeline JSON in the GUI — a great idea, but in practice 1.5B is too weak for non-trivial pipelines. I connected Claude Code via duckle-mcp, which works beautifully for creating complex pipelines (multi-stage unpivot → regex extract → pivot chains, etc.), but then I hit the GUI sync wall described below.

## Problem 1: No bidirectional GUI ↔ JSON sync

**Current behavior:**
- When the GUI edits a pipeline, it saves to the JSON file immediately ✓
- When the JSON file is modified externally (by MCP, git pull, hand-editing), the GUI **does not pick up the change** — even after switching to another pipeline and back ✗
- If the GUI has a pipeline open in memory and the JSON file is changed, the next GUI save **silently overwrites the external changes** ✗

**Steps to reproduce:**

1. Open a pipeline in the Duckle GUI (e.g. `dsd_drug_export`)
2. Use duckle-mcp `create_pipeline --overwrite` to modify the same pipeline's output path
3. Switch to another pipeline in the GUI, then switch back
4. The GUI still shows the old output path
5. Make any edit in the GUI and save — the MCP changes are overwritten

**Expected behavior:**
- At minimum: **File → Reload from Disk** menu item
- Ideally: a file-system watcher that detects external JSON changes and prompts "This file was modified externally. Reload?"

This is the single biggest friction point for a "git-friendly, file-based" workspace — the files should be the source of truth, not the GUI's in-memory state.

## Problem 2: MCP lacks a proper `update_pipeline` tool

**Current behavior:**
- `create_pipeline` can overwrite, but requires the **entire** pipeline object
- This means the caller must first `read_pipeline`, merge changes, then `create_pipeline --overwrite`
- MCP-created pipelines do **not** auto-register in `repository.json` or `duckle.json`, so they are invisible to the GUI until the user manually edits those index files

**Expected behavior:**
- An `update_pipeline` tool that accepts a partial pipeline diff (e.g. only the changed node property) and merges it
- Or at minimum, `create_pipeline` should also write the corresponding entries into `repository.json` and `duckle.json` so new pipelines appear in the GUI automatically

## Problem 3: Built-in AI only supports the local Qwen model

[...]

PR #96, opened 2026-06-23
Feature/nodge config json

luyandamncube:
## Type

- [ ] Bug fix
- [x] New feature (connector / transform / sink)
- [ ] Performance
- [ ] Docs
- [ ] Refactor / internal

## Checklist

- [ ] `cargo fmt` and `cargo clippy --workspace --all-targets -- -D warnings` pass
- [ ] `cargo test --workspace` passes (engine tests need `DUCKLE_DUCKDB_BIN` set - see CONTRIBUTING.md)
- [ ] Commits are small with imperative subject lines
- [ ] No code ported from incompatibly licensed sources


SouravRoy-ETL:
Thanks for the PR @luyandamncube, there's some genuinely useful work in here (the per-node JSON config tab is clean, and the HTTP transport idea is reasonable). Before we can take any of it though, a few things need sorting out, grouped by priority so it's easy to work through.

**First, a quick intent check.** A lot of this PR is branded "Stitchly" (around 58 references, plus a `.stitchly/` runtime temp dir), and the new docs under `docs/00_foundation/` frame Duckle as a fork to be rebranded into "Stitchly v2". If this branch was actually meant for a downstream fork of yours rather than Duckle itself, no worries, just let me know. If it's meant for Duckle, all the Stitchly naming and the fork/rebrand framing needs to come out, since this repo is its own clean-room project.

**Blocking (security):**
- The new HTTP bridge sets `Access-Control-Allow-Origin: *` on endpoints that run arbitrary pipelines with no auth. Because a pipeline can contain `code.shell`, any website a user visits while the server is running could POST to localhost and get code execution; the autodetect endpoint similarly lets any origin read arbitrary files off the host. This needs to be restricted to a same-origin / localhost allowlist (ideally plus a token) before it can land.
- `workspacePath` is read straight from the request body, which bypasses the workspace-containment check the existing endpoints enforce.

**Blocking (mergeability + hygiene):**
- The branch is ~52 commits behind `main` and currently conflicts in 5 files (`crates/duckdb-engine/src/lib.rs`, `plan/mod.rs`, `plan/tests.rs`, `crates/duckle-runner/src/main.rs`, `frontend/src/tauri-bridge.ts`). A large part of the +17k/-4.7k diff is just rustfmt reflow against the old base. Please `git rebase` onto current `main` and re-push, it should shrink the diff a lot and resolve the conflicts against the QVD, GizmoSQL and #92/#93 work that has since landed.
- `runs/p_mqqc4kkt_92cno.json` and `runs/p_mqqdcfl6_5fuy6.json` are generated run-output artifacts. Please drop them and add `runs/` to `.gitignore`.

**Please split this up.** Right now it bundles roughly 5 separate things (the JSON config tab, the HTTP bridge, the `code.shell` upstream-input change, the Dolt scripts, and a ~40-file docs set). That is hard to review and risky to merge as one unit. Smaller focused PRs would move much faster, roughly: (1) JSON config tab, (2) `code.shell` upstream input, (3) HTTP bridge after the CORS fix, (4) Dolt, (5) docs.

**Dolt scripts:** these shell out to the `dolt` CLI but need `set -o pipefail` (right now a failed `dolt sql` / `dolt ls` inside a pipe is silently swallowed and can produce wrong data), and the table name is interpolated unquoted into both `dolt sql` and the DuckDB identifier queries (injection risk). Best landed as their own PR once de-branded.

**Last thing:** the PR description is still the empty template. Could you fill in what it does, link any related issue, and tick the checklist (fmt / clippy / tests)?

Happy to review each split PR quickly once the above is sorted. Thanks again for digging in.

Issue #170, opened 2026-07-13
REST source with 0 records materializes a single `json` column, breaking downstream SQL


angusmaul:
A REST-family source (hit via `src.salesforce`, but the mechanism is shared) that receives an
empty result set produces a relation whose only column is a JSON-typed `json`, so the next
node's SQL fails with a confusing binder error instead of seeing an empty typed relation.

**Repro**
1. `src.salesforce` query whose SOQL matches nothing (e.g. `WHERE External_ID__c LIKE 'NOMATCH-%'`),
   `responsePath /records`
2. → `code.sql` with `SELECT Id FROM input`
3. Run: source reports `ok (0 rows)`, then the SQL node fails:
   `Binder Error: Referenced column "Id" not found in FROM clause! Candidate bindings: "json"`

With ≥1 record returned the same pipeline works.

**Cause**
`JsonLinesWriter::finalize_into_table` materializes the node table via
`read_json_auto('<tmp>.ndjson', format='newline_delimited', sample_size=-1)`. With zero rows
the NDJSON file is empty, and `read_json_auto` over an empty file yields a single `json`
column — so every downstream column reference breaks. Affects all `JsonLinesWriter` /
`materialize_jsonobjects_as_table` consumers.

**Expected**
An empty result should produce an empty relation with the right columns — e.g. from the node's
declared schema when one is present — or failing that, a clear source-level message
("<node>: query returned 0 records and no schema is declared to type an empty result"),
rather than a binder error surfacing on the next node.

Found while building the live test suite for #166 stage 2 (running the delete step twice —
second run's verification query legitimately returns 0 records). Pre-existing on v0.5.5,
unrelated to the #166 changes.

SouravRoy-ETL:
Fixed on main in 0a854ad.

A result with 0 rows no longer materializes a single `json` column that breaks downstream SQL. The fix lives at the single point where every JSON-materializing consumer finalizes (about 50 of them: the REST source in your report, plus email, the AI/code transforms, and the rest), so it applies uniformly. When the node has a declared schema the empty result is typed from it; transforms mirror their upstream schema; otherwise you get a clear error instead of a misleading `json` column.

This will be included in the next release, planned for 18 July 2026.

LouisDeconinck:
The zero-row materialization bug was fixed centrally in 0a854ad for REST and the other JSON-materializing components. Can this issue be closed?

LouisDeconinck:
@angusmaul can you close this issue please?

Issue #164, opened 2026-07-11
Feature Request: Salesforce sink (snk.salesforce) for write-back / migration into Salesforce


angusmaul:
## The problem

Salesforce is currently a **source only** (`src.salesforce`, a thin wrapper over `src.rest`). There is no Salesforce **sink**, and in fact no SaaS application has a write sink today — the 66 sinks are files, databases, warehouses, object stores, streaming, NoSQL, and vector DBs.

That makes Duckle a great fit for *reading* Salesforce into a warehouse/lake, but it can't support the other half of real Salesforce work: **ETL and data migration *into* Salesforce** (seeding a new org, syncing records back, ongoing reverse-ETL from a warehouse to Salesforce).

The only way to write to Salesforce today is the generic `snk.rest` / `snk.webhook` / `snk.graphql` path pointed at the Salesforce REST API, which has real limits:

- `snk.webhook` = one HTTP call per record → slow and burns the daily API limit fast at migration volume.
- `snk.rest` (batch-into-array) can't produce the shapes the Salesforce Composite/sObject Collections API requires: the `{ "allOrNone": …, "records": [ … ] }` wrapper and the per-record `attributes: { "type": "Account" }` envelope aren't expressible.
- No **Bulk API 2.0** (the correct tool for migration volume — it's a multi-step job lifecycle, not a single POST).
- No **upsert by external ID**, no partial-success / error-file handling, no parent→child **ID remapping**. Those are exactly the things a Salesforce migration lives or dies on.

## Proposed solution

A first-class **`snk.salesforce`** sink covering the two Salesforce write APIs, chosen by volume:

- **Bulk API 2.0** (`/services/data/vXX.0/jobs/ingest`) for migration-scale loads — create job → upload CSV → close → poll status → fetch success/failed result sets.
- **sObject Collections / Composite** (`/composite/sobjects`, up to 200 records/call) for smaller batches and lower latency.

Suggested config surface:

- `object` (sObject API name, e.g. `Account`), `apiVersion`
- `operation`: `insert` | `update` | `upsert` | `delete`
- `externalIdField` (required when `operation = upsert`)
- `api`: `bulk` | `rest` (or `auto` by row count)
- `allOrNone` / `batchSize`
- Auth: Bearer OAuth token (same as `src.salesforce`), via `${ENV:...}` so no secrets in the pipeline JSON
[...]

PR #196, opened 2026-07-20
feat(salesforce): #164 src.salesforce.bulk - Bulk API 2.0 query source for migration-scale reads

angusmaul:
Adds **`src.salesforce.bulk`** — the Bulk API 2.0 **query source**, completing the migration-scale pair with the merged sink (#187). A SOQL statement runs as an async query job (`query` / `queryAll` — the latter includes deleted and archived records), and the paged CSV result sets stream to disk before DuckDB materializes them, so a multi-GB extract never lands in memory on either leg.

## Design notes

**1. Shares the sink's machinery, not `src.salesforce`'s.** The REST-form `src.salesforce` is a thin alias over `src.rest` (user-authored URL, `authType`/`authToken` keys, ~2k records per page). The Bulk source owns its endpoint and drives the same job lifecycle as the Bulk sink, so it takes the sink-shaped auth keys (`authMode`/`instanceUrl`/`accessToken`), the same client-credentials mint-per-run, and the same saved-connection resolution — the `duckle-secrets` gate has recognised `src.salesforce.bulk` since #187, which is also what keeps org-A-read → org-B-write working across the pair. `bulk_poll_ingest_job` is generalised to `bulk_poll_job` (interval/timeout as params) so both job families share the cancellable, abort-on-timeout poller.

**2. `Sforce-Locator` pagination, streamed.** Result pages walk `GET /jobs/query/{id}/results` (optional `maxRecords` page size); the next page's handle arrives in the `Sforce-Locator` response header, and the last page is signalled by the **literal string `"null"`** — not an absent header — which is called out in a comment since it's an easy thing to "fix" into a bug. Pages append through the same per-file-header logic your #187 follow-up introduced for the sink's result files, so the staging file carries exactly one header. Staging uses the same owner-only (`0700`) temp-dir pattern.

**3. The #170 typed-empty contract, now with a regression test.** A 0-record query with a declared node schema materializes a typed empty relation via `materialize_empty_result`; without one it fails with the clear source-level error your #170 fix established. This PR adds the first e2e regression for that contract (the fix shipped without one). With rows and a declared schema, columns are pinned via `all_varchar` + `TRY_CAST` — a stray unparseable cell becomes NULL rather than failing a multi-GB load.

**4. SOQL restrictions surface as Salesforce's own message.** Bulk 2.0 rejects GROUP BY, OFFSET, TYPEOF, aggregates and parent-to-child subqueries at job creation; the API's `MALFORMED_QUERY` message lands in the run error verbatim.

## Validation

- **Mock:** 4 e2e tests on a stateful query-job mock — locator pagination lands 2 pages' rows in order with one header; 0-record + declared schema flows through downstream SQL as a typed empty relation; a `Failed` job surfaces the API message; a poll timeout fails the run and PATCHes the job to `Aborted`. Full engine suite green (310 unit + 274 e2e).
- **Live suite:** two new steps — q1 runs the `BULK-*` SOQL as a real query job while the records exist (4 rows incl. the upserted one, through the saved `${ENV:}`-placeholder connection); q2 reruns it on the clean org and asserts the typed-empty, header-only CSV. **31/31 green** against a real org (27 existing + 4 new).

Frontend: `Salesforce Bulk` source entry in the CRM group, manifest with the saved-connection picker / `visibleWhen`-gated inline auth / SOQL textarea / `query|queryAll` / `maxRecords` / poll+timeout knobs, `schemaSource: 'declared'`. Catalog regenerated from the freshly-fixed baseline (375 components) — purely additive this time.


SouravRoy-ETL:
Merged. This is a strong PR: the Bulk source correctly reuses the sink's job machinery rather than `src.salesforce`'s REST shape, the poller generalisation leaves the shipped sink path byte-identical (same two values, same order), the locator walk terminates on the literal `"null"` as well as on an empty or absent header, the per-file header logic keeps exactly one CSV header across pages, and the catalog was regenerated against the fixed exporter so it is genuinely additive. The `queryAll` variant and the #170 typed-empty regression test are both welcome.

One thing I fixed on main straight after merging, in c6b9a2f.

Emptiness was decided from `Sforce-NumberOfRecords` alone:

```rust
total_rows += resp.header("Sforce-NumberOfRecords")
    .and_then(|s| s.parse::<u64>().ok())
    .unwrap_or(0);
```

Every page is streamed to the staging file regardless, but that header then gated the empty branch. `Sforce-*` headers are non-standard and optional, so an egress proxy or gateway that strips or rewrites them leaves `total_rows` at 0 with a complete result set already on disk. The run would take the empty branch, discard it, and report `0 rows (typed empty relation)` as a success.

That is the one failure mode a migration source must not have: the pipeline goes green and the target is simply empty, with nothing in the logs to suggest otherwise.

The staged file is now the authority and the header is a fast path. `result_csv_has_data_rows` stops at the first non-blank line after the header, so it is free even on a multi-GB file, and a header-only result still takes the typed-empty path. When the header is missing but rows are staged, that discrepancy now prints, since it means something upstream is rewriting responses.

Not blocking, but worth picking up if you continue on this connector:

- **Locator encoding.** The locator is interpolated raw into the results URL, while the REST cursor pagination in the same file passes its token through `urlencode_simple`. A locator containing `+` would decode to a space server-side.
- **Runaway guard.** The page walk is an unbounded `loop` with no page cap and no check that the locator advanced, and `timeoutSecs` bounds only the poll phase. A peer that returns a non-advancing locator would append the same page forever.
- **No-declared-schema typing.** Without a declared schema the read falls to `read_csv` inference, so a `BillingPostalCode` of `01234` sniffs as BIGINT and loses its leading zero. The declared-schema branch pins `all_varchar` + `TRY_CAST` and does not have this problem, so it may be worth defaulting the source to text-preserving reads.
- **Staging path.** `duckle-sfbulkq-<pid>-<seq>` in the shared temp dir is predictable; `create_private_dir` sets the mode but accepts a pre-existing directory.

Thanks for this, and for the live-suite steps. Ships in the next release.

SouravRoy-ETL:
Two follow-ups now that the full review finished.

**The header issue was reproduced, and it is sharper than I described above.** Patching the mock to omit only `Sforce-NumberOfRecords`, leaving `Sforce-Locator: null` and the body intact, gives:

```
status=ok, error=None
output CSV: "Id,Name" only, line_count=1
```

Two records that arrived on the wire and were written to the staging file, gone, with the run reporting success. The part I got wrong earlier: this is silent **specifically in the declared-schema configuration**, which is the one the manifest steers users to (`schemaSource: 'declared'`) and the one your `q1_bulk_query.json` fixture uses. Without a declared schema `materialize_empty_result` errors, so the safest-looking setup was the vulnerable one. Fixed in c6b9a2f.

**Locator encoding is done too** (c4e9ba7), since it turned out to be one call and an existing convention rather than a new one: this file already runs the `src.rest` cursor token and Weaviate's `after` cursor through `urlencode_simple`. `urlencode_simple` leaves alphanumerics and `-_.~` untouched, so every locator seen in practice serializes byte-identically and your pagination test is unaffected.

Everything else in the review came back clean, including the two I most wanted checked: the poller generalisation leaves the shipped sink byte-identical (same two values, same order, body unchanged, all 29 pre-existing Salesforce tests pass), and the catalog regenerates byte-identically against the fixed exporter, so it does not revert that change.

Still yours if you want them, all non-blocking: the unbounded page loop (no locator-progress check, and `timeoutSecs` bounds only the poll phase), and the no-declared-schema read falling to `read_csv` inference so a `01234` postcode sniffs to BIGINT.

angusmaul:
Thanks for the thorough review, and for the two fixes on main — the staged-file-as-authority change especially. You are right that a silent 0-row success is the one outcome a migration source must never produce, and gating the empty branch on an optional header was exactly that failure mode; reproducing it through the mock with only Sforce-NumberOfRecords stripped makes it unambiguous. Same lesson as the sink round's 10 MB into_string discovery: trust the artifact, not the transport signal.

We'll pick up both remaining items in a fast-follow PR: the runaway guard on the results page walk (fail on a non-advancing locator, plus a sensible overall ceiling so timeoutSecs isn't the only bound), and defaulting the no-declared-schema read to all_varchar so values like a 01234 postcode survive as text and any casting is a deliberate downstream choice.

PR #197, opened 2026-07-21
fix(salesforce): #164 Bulk query result-walk guards + text-preserving no-schema reads

angusmaul:
The fast-follow promised on #196, covering both remaining review items:

**Runaway guards on the result page walk.** `timeoutSecs` bounds only the poll phase, so the walk now carries its own: a non-advancing `Sforce-Locator` (a peer or middlebox echoing the same token back) fails the run naming the job and the stuck locator, instead of re-appending the same page until the disk fills. A far-out page ceiling (50M pages — commented as a runaway backstop, not a tunable) additionally bounds a peer that keeps minting fresh locators.

**Text-preserving no-schema reads.** Without a declared schema the read now passes `all_varchar` to `read_csv` rather than letting inference sniff types — a `BillingPostalCode` of `01234` sniffed to BIGINT `1234`, silently dropping the leading zero. Salesforce serves every value as CSV text anyway; casting becomes a deliberate downstream choice, and a declared schema still pins types via `TRY_CAST` exactly as before.

Tests: a new mock e2e drives the non-advancing-locator failure (the query mock gained a stuck-locator mode), and the pagination e2e now also asserts leading-zero survival — an assertion that fails without the `all_varchar` change. 5 source e2e + the full Salesforce suite (17 unit + 19 e2e, including your two post-#196 fixes' tests) green. IMPLEMENTATION.md updated for both behaviours.

### DIFF
--- a/crates/duckdb-engine/src/connectors.rs
+++ b/crates/duckdb-engine/src/connectors.rs
@@ -1026,6 +1026,7 @@ impl DuckdbEngine {
         // edge of the API.
         let mut total_rows: u64 = 0;
         let mut locator: Option<String> = None;
+        let mut pages_fetched: u64 = 0;
         loop {
             self.check_cancelled()?;
             let mut url = format!("{}/{}/results", query_base, job_id);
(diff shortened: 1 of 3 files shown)


SouravRoy-ETL:
Merged. Both guards are exactly right, and the non-advancing-locator check is the one that actually matters: it is precise (first page is safe, since `locator` is `None` and cannot match), it names the job and the stuck token, and it fails instead of filling the disk. Verified locally on a trial merge with a real DuckDB before merging: 311 unit and 275 e2e green, and all 10 Salesforce bulk e2e genuinely executed rather than skipping.

One correction pushed on top, in 7e8a1c4.

The `all_varchar` change is right, but not for the stated reason, and its test did not prove it. DuckDB 1.5.4's sniffer already keeps `01234` as VARCHAR, so inference was never turning that postcode into `1234`. I reverted only the `all_varchar` argument and the leading-zero assertion still passed, which makes it a test that cannot fail for the reason it exists. Given the session we have both just had on silent failures, a green assertion that proves nothing felt worth fixing rather than leaving.

The real hazard is type instability between extracts, and it is a good one. On 1.5.4:

```
ExtId all numeric          -> BIGINT
same column, one alphanum  -> VARCHAR
```

So the same Salesforce column changes type depending on what a given extract happened to contain, which for a migration source means the target table changes shape between runs or the load fails on a type mismatch. `all_varchar` removes that, and lines the source up with the declared-schema branch and the `src.xml` / Snowflake / Teradata typed-finalize paths.

The fixture now carries an `Amt` column with trailing-zero decimals, which is a case where inference IS visible in the output: sniffed to DOUBLE, `2.50` comes back as `2.5` and `10.00` as `10.0`. I confirmed the assertion fails without `all_varchar` and passes with it. The leading-zero check stays as a weaker companion.

The `BULK_QUERY_MAX_PAGES` framing as a backstop rather than a tunable is the right call, and the comment explaining why the locator check is the real guard is worth keeping.

That closes out both review items from #196. Thanks for turning these around quickly, and for the mock's stuck-locator mode, which is what made the guard testable at all.

PR #199, opened 2026-07-21
Fix Python pipeline limit serialization

### LINKED ISSUE #198: Python Pipeline.limit(n) always compiles to LIMIT 100
## What happened

The Python API's `Pipeline.limit(n)` ignores the requested value at compile time. It serializes the node as `{"count": n}`, but the Rust `xf.limit` builder only reads `limit` or `rows`. Because unknown properties are ignored, the pipeline validates successfully and silently falls back to `LIMIT 100`.

This means `.limit(7)` allows up to 100 rows through, while `.limit(1000)` silently drops everything after the first 100 rows.

## Steps to reproduce

1. Install Duckle 0.5.9.
2. Run:

```python
import duckle

p = duckle.read_csv("input.csv").limit(7)
print(p.to_dict()["nodes"][-1]["data"]["properties"])
p.save("limit.json")
```

The printed properties are:

```python
{'count': 7}
```

3. Compile without opening the source or running a sink:

```sh
duckle validate --json --sql limit.json
```

The command exits successfully, but the generated stage contains:

```sql
SELECT * FROM "csv" LIMIT 100
```

For comparison, changing the node property in `limit.json` from `{"count": 7}` to `{"limit": 7}` produces `LIMIT 7`.

## Expected behavior

`Pipeline.limit(7)` should serialize the engine-supported property and compile to:

```sql
SELECT * FROM "csv" LIMIT 7
```

## Environment

- Duckle/PyPI version: 0.5.9
- Verified on: macOS arm64
- The mismatch is platform-independent: `packaging/pypi/duckle/api.py` emits `count`, while `crates/duckdb-engine/src/plan/builders.rs::build_limit` reads `limit` or `rows`.

## Suggested fix

Change the Python mapping from:

```python
{"count": int(n)}
```

to:

```python
{"limit": int(n)}
```

and add a focused regression test asserting that `Pipeline().limit(7)` serializes `{"limit": 7}`.

mattfaltyn:
## What this changes

Serialize Python `Pipeline.limit(n)` nodes with the engine-supported `limit` property instead of `count`. This prevents the compiler from silently falling back to `LIMIT 100`, and adds a focused regression test for the Python-to-engine contract.

## Related issue

Fixes #198

## Type

- [x] Bug fix
- [ ] Performance
- [ ] Docs
- [ ] Refactor / internal

## Checklist

- [ ] `cargo fmt` and `cargo clippy --workspace --all-targets -- -D warnings` pass

## Notes for reviewers

Validated locally with:

- `PYTHONPATH=packaging/pypi python3 -m unittest discover -s packaging/pypi/tests -v`
- an end-to-end Python API compile assertion confirming `.limit(7)` produces `LIMIT 7`
- `cargo test --workspace`
- `git diff --check`

The frontend was not changed, so its lint command is not applicable. On the current upstream revision, the Rust 1.93 formatter and clippy commands report existing failures in untouched Rust files (including `crates/connectors/src/csv.rs` and `crates/duckdb-engine/tests/execution.rs`); this PR does not modify Rust code.

### DIFF
--- a/packaging/pypi/duckle/api.py
+++ b/packaging/pypi/duckle/api.py
@@ -182,7 +182,7 @@ class Pipeline:
         return self._add("transform", "xf.sort", {"orderBy": order}, "Sort")
 
     def limit(self, n):
-        return self._add("transform", "xf.limit", {"count": int(n)}, "Limit")
+        return self._add("transform", "xf.limit", {"limit": int(n)}, "Limit")
 
     def dedupe(self, *columns):
         return self._add("transform", "xf.distinct", {"columns": list(columns)}, "Dedupe")
--- a/packaging/pypi/tests/test_api.py
+++ b/packaging/pypi/tests/test_api.py
@@ -1,0 +1,16 @@
+import unittest
+
+import duckle
+
+
+class PipelineTests(unittest.TestCase):
+    def test_limit_uses_engine_property_name(self):
+        pipeline = duckle.read_csv("input.csv").limit(7)
+
+        properties = pipeline.to_dict()["nodes"][-1]["data"]["properties"]
+
+        self.assertEqual(properties, {"limit": 7})
+
+
+if __name__ == "__main__":
+    unittest.main()


SouravRoy-ETL:
Merged. Good catch, and the diagnosis was exactly right: `build_limit` reads `props["limit"]` with a `rows` fallback and otherwise defaults to 100, so `count` was never seen.

I reproduced it before merging rather than take it on faith. Against a 4-row CSV, `.limit(2)` returned all 4 rows, because the unrecognised key fell through to `LIMIT 100`. With your change it returns 2. That silence is the part that makes it nasty: nothing errors, the status is green, and the only symptom is a number that happens to be wrong.

Because of that, I checked whether the rest of the curated API had the same problem. It did, in two more places, both pushed in 9276f75:

- **`read_duckdb`** wrote `path` / `table`, but the engine reads `database` (via `attach_prelude`) and `tableName`. Nothing was attached, so a run against a real 4-row DuckDB table returned **zero rows and reported ok**. Worse than the limit case, same cause.
- **`select`** emitted `xf.select`, which is not a component. `xf.project` is the one labelled "Project / Select". That one at least failed loudly at compile time.
- **`read_postgres`** wrote `table` where the engine reads `tableName`.

Two things about the test you added, since both are worth knowing.

It was the repo's first Python test, and **nothing ran it** - there was no CI job that would execute `packaging/pypi/tests`. A regression test that never runs cannot prevent a regression, so there is now a `python-api` job in ci.yml, and the file covers every curated method's component id and property keys rather than just `limit`.

Writing that out also surfaced something neither of us was looking for: `xf.limit` and `xf.pyexpr` are executable but absent from the palette, so the catalog does not list them and `list_components` cannot see them. `.derive()` runs on `xf.pyexpr`, so a headline feature of the Python API is invisible to the GUI and to agents. Named explicitly in the test so the list cannot grow quietly; I will fix the catalog gap separately.

On CI not running here: the PR is from a fork, so the workflows needed approval and none had been granted. Nothing to do with your branch.

Thanks for this. Ships in the next PyPI release.

PR #275, opened 2026-08-27
fix(workspace): support inline duckle.json rollups and robust active job recovery (#213)

### LINKED ISSUE #213: Duckle desktop opens valid workspace but shows starter orders_etl project
## What happened

Duckle desktop appears to open a selected workspace folder, but the Project panel still shows the bundled starter `orders_etl` project instead of the jobs from the selected workspace.

Expected instead:

- The app should show active job `workspace_coverage`.
- The app should load the portfolio jobs from `<workspace-root>/runtime/duckle-portfolio-rollup/duckle.json` or the compatibility `<workspace-root>/runtime/duckle-portfolio-rollup/workspace.json`.

Observed:

- After opening `<workspace-root>/runtime/duckle-portfolio-rollup`, the app still shows `orders_etl`.
- This still happens after upgrading Duckle desktop to build 0.5.8.
- The same workspace validates locally by CLI.

## Steps to reproduce

1. From a local workspace containing a generated Duckle rollup, run `npm run duckle:all:verify`.
2. Launch Duckle desktop with `tools/duckle/bin/duckle`.
3. Open `<workspace-root>/runtime/duckle-portfolio-rollup`.
4. Check the Project panel.
5. Expected `workspace_coverage`; observed `orders_etl`.

## Pipeline / SQL

No SQL query is involved. This looks like a desktop workspace-loader/cache/starter fallback issue.

Local workspace evidence:

- `duckle.json` has 62 jobs and `activeJobId: "workspace_coverage"`.
- `workspace.json` has 62 jobs and 62 `pipelineData` entries.
- The rollup folder has 193 files, including 124 pipeline files and 63 data files.
- `npm run duckle:all:verify` passes 29/29 local Duckle verifiers.
- `npm run duckle:portfolio:verify` passes.

Key local file hashes from the sanitized repro packet:

- `runtime/duckle-portfolio-rollup/duckle.json` sha256 `61337efa3e172d90116a05d13ad2cdb5a8a4c8986ebf100cd1fe912dc3fd477b`
- `runtime/duckle-portfolio-rollup/workspace.json` sha256 `893500cec9a2d77658629500525d5c3e3a62c9a791a138e698a92bde729d0007`
- `runtime/duckle-portfolio-rollup/repository.json` sha256 `ffb421771761cb61f09efa2a5b3cf92224deea0bc071721a8a0423b01a474a2a`

## Environment

- Duckle version / release tag: desktop build 0.5.8, per in-app upgrade confirmation
- OS + architecture: Ubuntu / Linux x86_64 environment with GUI display available
- For a database connector: not applicable

## Logs / error text

No app error text was shown in the UI. The failure mode is that the desktop app displays `orders_etl` after selecting a valid workspace directory whose local JSON files contain `workspace_coverage` and 62 jobs.

This report includes only sanitized local repro details and file hashes; it does not include private account material or customer records.

FrancescoCastaldi:
﻿### Description
Fixes #213

This PR resolves an issue where opening valid workspaces (such as CLI-generated rollups, single-file exports, or workspaces where pipelines/jobs are in `duckle.json` or `workspace.json`) would fail to hydrate the active job list and incorrectly fall back to displaying the starter `orders_etl` project.

### Root Cause
1. `loadV2` in `frontend/src/workspace.ts` ignored inline `pipelineData` and `repo` embedded in `duckle.json`, and only looked for items in `repository.json` and `pipelines/*.json`.
2. When `duckle.json` lacked an explicit `jobs` array or separate `repository.json`, `jobs` stayed `undefined` and `pipelineData` remained empty, leaving `App.tsx` on the default `orders_etl` state.
3. If `duckle.json` was minimal/empty while `workspace.json` held the actual jobs/pipelines, `loadV2` did not fall back to `workspace.json`.

### Changes
- **`frontend/src/workspace.ts`**:
  - Support reading inline `pipelineData` and `repo` from `duckle.json`.
  - Scan the `pipelines/` folder for any on-disk pipelines not yet in `repository.json` and auto-register them in `repo`.
  - Dynamically derive `jobs` and `activeJobId` from `repo` or `pipelineData` when `meta.jobs` is not provided.
  - Add seamless fallback hydration from `workspace.json` if `duckle.json` contains no jobs or pipelines.
- **`frontend/src/App.tsx`**:
  - Ensure `setActiveJobId` uses the loaded `state.activeJobId` or the first loaded job rather than keeping the default starter job ID.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -589,9 +589,9 @@ export default function App() {
                 }
                 if (state) {
                     if (state.engine) setEngine(normalizeEngineId(state.engine));
-                    if (state.pipelineData && Object.keys(state.pipelineData).length > 0)
+                    if (state.pipelineData)
                         setPipelineData(state.pipelineData as Record<string, PipelineState>);
-                    if (state.repo && (state.repo as RepoItem[]).length > 0)
+                    if (state.repo)
                         setRepo(state.repo as RepoItem[]);
                     if (state.jobs && (state.jobs as Job[]).length > 0) {
                         const loadedJobs = state.jobs as Job[];
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -589,11 +589,21 @@ export default function App() {
                 }
                 if (state) {
                     if (state.engine) setEngine(normalizeEngineId(state.engine));
-                    if (state.pipelineData)
+                    if (state.pipelineData && Object.keys(state.pipelineData).length > 0)
                         setPipelineData(state.pipelineData as Record<string, PipelineState>);
-                    if (state.repo) setRepo(state.repo as RepoItem[]);
-                    if (state.jobs) setJobs(state.jobs as Job[]);
-                    if (state.activeJobId) setActiveJobId(state.activeJobId);
+                    if (state.repo && (state.repo as RepoItem[]).length > 0)
+                        setRepo(state.repo as RepoItem[]);
+                    if (state.jobs && (state.jobs as Job[]).length > 0) {
+                        const loadedJobs = state.jobs as Job[];
+                        setJobs(loadedJobs);
+                        const targetActiveId =
+                            state.activeJobId && loadedJobs.some(j => j.id === state.activeJobId)
+                                ? state.activeJobId
+                                : loadedJobs[0].id;
+                        setActiveJobId(targetActiveId);
+                    } else if (state.activeJobId) {
+                        setActiveJobId(state.activeJobId);
+                    }
                     if (state.corruptFiles?.length) setCorruptFiles(state.corruptFiles);
                 }
                 setWorkspaceReady(true);
--- a/frontend/src/workspace.ts
+++ b/frontend/src/workspace.ts
@@ -297,7 +297,9 @@ async function loadV2(path: string): Promise<WorkspaceState | null> {
         }
     }
 
-    // Scan pipelines/ directory for any pipeline files on disk not explicitly referenced in repo
+    // Auto-register any pipeline files present on disk in `pipelines/` that are
+    // missing from repository.json. This is intentional for workspace recovery
+    // (e.g. from git merges or rollups where repository.json is partial).
     const diskPipelineFiles = await readDirEntries(joinPath(path, PIPELINES_DIR));
     for (const pfile of diskPipelineFiles) {
         const pid = pfile.replace(/\.json$/i, '');
--- a/frontend/src/workspace.ts
+++ b/frontend/src/workspace.ts
@@ -245,12 +245,14 @@ async function loadV2(path: string): Promise<WorkspaceState | null> {
         engine?: string;
         jobs?: unknown[];
         activeJobId?: string;
+        pipelineData?: Record<string, unknown>;
+        repo?: unknown[];
     }>(joinPath(path, METADATA_FILE));
     if (!meta) return null;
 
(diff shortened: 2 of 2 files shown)


FrancescoCastaldi:
Hi @SouravRoy-ETL @brightleaddigital! 👋 This PR fixes #213 by adding robust support for loading inline workspace rollups/single-file duckle.json, disk pipeline scanning, and fallback hydration from workspace.json. Ready for review!

SouravRoy-ETL:
Thanks for this - the diagnosis in `loadV2` is right and the workspace.ts side is the good half. Reading inline `pipelineData`/`repo` from `duckle.json`, deriving `jobs` from the repo when `meta.jobs` is absent, and picking `activeJobId` from the loaded jobs rather than leaving the starter selected all address #213 directly. It builds clean.

**One change I do not think we can take as written**, and it is in App.tsx rather than the loader.

```ts
if (state.pipelineData && Object.keys(state.pipelineData).length > 0)
    setPipelineData(...)
if (state.repo && (state.repo as RepoItem[]).length > 0)
    setRepo(...)
```

An empty object is truthy, so the old code DID call `setPipelineData({})` and that is what clears the previous workspace out of memory. With the length guard, an empty load no longer clears anything.

That matters because switching workspaces happens **in place**. The hydrate effect is keyed on `[workspacePathState, reloadNonce]` (App.tsx:625) and `setWorkspacePathState(picked)` at App.tsx:861 changes it without a remount or a reload - the only `window.location.reload()` is the explicit button at :2617. The effect also never sets `workspaceReady` back to false; it only ever sets it to `true` at :609.

So:

```
open workspace A  (3 pipelines)   -> pipelineData = {A...}
switch to workspace B (genuinely empty)
  loadV2(B) returns pipelineData {}
  guard skips setPipelineData        <- A's pipelines stay in memory
  setWorkspaceReady(true) at :609
  autosave writes A's pipelines into B
```

That is the failure the comment at App.tsx:415-417 is guarding against - *"stay un-ready until the workspace has hydrated, so auto-save can't overwrite the server/disk files with empty in-memory defaults"* - arriving from the other direction: non-empty defaults over an empty workspace.

**The good news is the loader change makes those guards unnecessary.** Now that `loadV2` derives `jobs`, `repo` and `pipelineData` from whatever it can find, an empty result means the workspace really is empty, and clearing is the correct response. So I would keep the `activeJobId` selection exactly as you wrote it and drop the two `length > 0` guards:

```ts
if (state.pipelineData) setPipelineData(...)
if (state.repo) setRepo(...)
if (state.jobs && (state.jobs as Job[]).length > 0) {
    ...your activeJobId logic, unchanged...
}
```

Two smaller notes, neither blocking:

- The `pipelines/` disk scan auto-registers every file it finds into `repo`. That is right for recovery, but it means a pipeline file the user deliberately removed from `repository.json` reappears on next load. Worth a comment saying that is intended, so nobody "fixes" it later.
- `repo` becomes `let` and is then mutated with `push`, while the items pushed carry only `{id,name,type,parentId}`. If `RepoItem` gains a required field later, these synthesised entries will be the ones that are missing it.

Happy to take it with the App.tsx guards dropped.

FrancescoCastaldi:
Good catch on in-place workspace switching @SouravRoy-ETL! 🎯 Updated to drop the length > 0 guards so setPipelineData and setRepo always execute and cleanly reset state on empty workspace loads. Also added the explanatory comment documenting the recovery intention for auto-registering disk pipelines.

PR #276, opened 2026-08-27
feat(frontend): support dynamic time offsets in template time builtins (#191)

### LINKED ISSUE #191: Support dynamic time offsets in template time builtins (src/rest.rs)
## The problem

Currently, the `insert_time_builtins` function in `src/rest.rs` only provides static time variables (`date`, `time`, `datetime`, `timestamp`, `now`) based on the exact moment the template is rendered.

There is no built-in way to generate **relative dates/times** (e.g. yesterday, tomorrow, 2 hours ago, etc.) directly in SQL templates or other templated content. This is a very common need when building scheduled pipelines, daily partitions, backtesting flows, or automated reports.

## Proposed solution

Extend `insert_time_builtins` to support **dynamic time offsets**.

**Desired behavior:**

- Support syntax such as `${date+1d}`, `${date-1d}`, `${now-2h}`, `${datetime+30m}`, etc.
- Accept an optional offset string in the function (e.g. `insert_time_builtins(vars, Some("1d"))`).
- Add a lightweight duration parser supporting days (`d`), hours (`h`), minutes (`m`), and seconds (`s`), with positive and negative signs.
- Keep full backward compatibility (no offset = current behavior).
- Optionally expose the offset via REST API (e.g. `?time_offset=1d` or `?time_offset=-2h`).

This would make time-related templating much more powerful and user-friendly.

## Alternatives / workarounds

- Manually calculating offsets in SQL using DuckDB date functions.
- Pre-computing variables in custom code before passing context to templates.
- Using external scripting to generate the variables.

These workarounds add complexity and reduce the convenience of the built-in template system.

## Additional context

- Related code: `src/rest.rs` (lines ~94-101)
- Common use cases: daily batch jobs, rolling window queries, report generation, testing with time shifts.
- Similar features exist in many template engines and ETL tools (e.g. Airflow macros, dbt date utils).

This enhancement would significantly improve the developer experience for time-sensitive pipelines in Duckle.

FrancescoCastaldi:
﻿### Description
Fixes #191

This PR brings full frontend parity to the dynamic time offsets feature implemented in the engine (`context.rs`), allowing templates, run resolution, and path substitutions to dynamically calculate relative time offsets.

### Features & Behavior
- **Dynamic Offset Syntax**: Support relative time expressions like `${date+1d}`, `${date-1d}`, `${now-2h}`, `${datetime+30m}`, etc.
- **Offset Parser**: Supports days (`d`), hours (`h`), minutes (`m`), and seconds (`s`) with signed offsets (`+`/`-`) and combined offsets (e.g. `+1d6h`).
- **Run Resolution & Parameter Discovery**:
  - `substituteString` and `substituteDeep` in `run-resolve.ts` now evaluate and substitute time builtins with offsets directly.
  - `discoverParams` accurately identifies offset time builtins as known builtins, preventing the UI from prompting users for manual values on relative time placeholders.

### DIFF
--- a/frontend/src/run-resolve.ts
+++ b/frontend/src/run-resolve.ts
@@ -208,14 +208,13 @@ export function resolveTimeBuiltin(name: string, now: Date = new Date()): string
  *   ${timestamp} -> epoch seconds
  *   ${now}       -> ISO-8601 (has colons; for values, not paths)
  */
-function timeBuiltins(): Record<string, string> {
-    const d = new Date();
+function timeBuiltins(now: Date = new Date()): Record<string, string> {
     return {
-        date: formatTimeBuiltin('date', d),
-        time: formatTimeBuiltin('time', d),
-        datetime: formatTimeBuiltin('datetime', d),
-        timestamp: formatTimeBuiltin('timestamp', d),
-        now: formatTimeBuiltin('now', d),
+        date: formatTimeBuiltin('date', now),
+        time: formatTimeBuiltin('time', now),
+        datetime: formatTimeBuiltin('datetime', now),
+        timestamp: formatTimeBuiltin('timestamp', now),
+        now: formatTimeBuiltin('now', now),
     };
 }
 
@@ -227,8 +226,8 @@ function timeBuiltins(): Record<string, string> {
  * separators are normalized to `/` (DuckDB accepts them on every platform).
  * The date/time builtins are always present, even without a workspace.
  */
-export function builtinVars(workspacePath?: string | null): Record<string, string> {
-    const vars = timeBuiltins();
+export function builtinVars(workspacePath?: string | null, now: Date = new Date()): Record<string, string> {
+    const vars = timeBuiltins(now);
     if (workspacePath) {
         const root = workspacePath.replace(/\\/g, '/');
         vars.workspace = root;
@@ -237,13 +236,13 @@ export function builtinVars(workspacePath?: string | null): Record<string, strin
     return vars;
 }
 
-function substituteString(value: string, vars: Record<string, string>): string {
+function substituteString(value: string, vars: Record<string, string>, now: Date = new Date()): string {
     return value.replace(/\$\{([^}]+)\}/g, (match, expr) => {
         const key = String(expr).trim();
         if (Object.prototype.hasOwnProperty.call(vars, key)) {
             return vars[key]!;
         }
-        const timeVal = resolveTimeBuiltin(key);
+        const timeVal = resolveTimeBuiltin(key, now);
         if (timeVal !== null) {
             return timeVal;
         }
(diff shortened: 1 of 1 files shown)


FrancescoCastaldi:
Hi @SouravRoy-ETL @jaty613! 👋 This PR completes frontend parity for #191 by adding dynamic time offset resolution and parameter discovery for \${date+1d}\, \${now-2h}\, \${datetime+30m}\, etc. Ready for review!

SouravRoy-ETL:
This is close to mergeable and the parity is the right instinct - a preview that resolves `${date+1d}` differently from the run is exactly the kind of divergence nobody notices until a path is wrong.

I checked it against the engine line by line and the important things agree:

- UTC on both sides (`chrono::Utc::now()` in context.rs:100/200, `getUTC*` here) - which matters, because a local-time frontend against a UTC engine would silently disagree by hours, and we have already been bitten by two schedulers disagreeing on timezone.
- Formats match exactly: `%Y-%m-%d`, `%H%M%S`, `%Y-%m-%d_%H%M%S`, epoch seconds, `%Y-%m-%dT%H:%M:%SZ` (context.rs:112-116).
- Base ordering `timestamp, datetime, date, time, now` matches context.rs:169, so `datetime` wins over `date` - getting that order wrong would make `${datetime}` resolve as `${date}` followed by a literal `time`.
- Offset parsing matches context.rs:126-156: optional sign, digits-then-unit, `d/h/m/s`, unknown unit rejected, trailing digits with no unit rejected, combined segments accumulated. A sign is required in both, because both only call the parser when `rest` starts with `+`/`-`.

**One thing I would change before merging.**

`substituteString` calls `resolveTimeBuiltin(key)` with its default argument, so a **fresh `new Date()` per placeholder**. The engine deliberately does not:

```rust
// One `now` for the whole pass so every placeholder (and every offset) in a
// run stamps the same instant.
let now = chrono::Utc::now();
```
context.rs:198-200

So in one resolution pass `${date}` comes from `timeBuiltins()`'s single `d`, while `${date+1d}` comes from a different instant taken microseconds later. Usually identical, and wrong exactly when it matters: a pass that straddles a second boundary gives `${time}` and `${time+1h}` an inconsistent base, and one straddling midnight gives `${date}` and `${date+1d}` a two-day gap instead of one.

The fix is small - thread one `Date` through, the way the engine threads one `now`:

```ts
function substituteString(value: string, vars: Record<string, string>, now = new Date()): string
```

and pass that same `now` into `resolveTimeBuiltin`, with `substituteDeep` creating it once per pass and `timeBuiltins()` taking it too, so the plain and offset builtins share an instant.

`discoverParams` treating an offset builtin as known rather than prompting for it is a real fix on its own - that would have been a prompt for `date+1d` on every run.

If you would rather I take the single-instant change on top after merge, that is fine too - say which you prefer.

FrancescoCastaldi:
Thanks for the thorough review @SouravRoy-ETL! 🚀 Updated to thread a single 
ow = new Date() instance across the entire resolution pass (
esolveForRun, uiltinVars, 	imeBuiltins, substituteDeep, substituteString, and 
esolveTimeBuiltin) so that both plain and offset builtins share the exact same instant, matching context.rs:198-200.

Issue #131, opened 2026-06-30
libodbc2 now required dependency on linux


add944:
Running the latest build 0.5.3 throws in the following error : 

"libodbc.so.2: cannot open shared object file: No such file or directory"

Installing the library solves the problem : "sudo apt install libodbc2"

it would be useful to include this in the installation instructions , thank you!

SouravRoy-ETL:
Thanks for the report @add944, and for digging up the fix.

You are right: the 0.5.3 Linux build picked up a hard dependency on `libodbc.so.2`. It comes from the Teradata ODBC connector being linked into the Linux binary, so the app needs the ODBC driver manager present just to launch. For right now, `sudo apt install libodbc2` (or `unixodbc` on non-Debian distros) gets you running, exactly as you found.

I will be re-rolling the Linux binary with a fix so this does not trip anyone up on a fresh machine, and I am adding a check to the release pipeline so a Linux build can never go out with this dependency again. I will update here once the new binary is on the release.

Appreciate the clear writeup.

SouravRoy-ETL:
The fixed Linux build is live now. I re-rolled the v0.5.3 Linux binaries so unixODBC is linked statically: the Teradata connector is still included, but the ODBC driver manager is baked into the binary, so it no longer needs libodbc.so.2 installed just to launch.

Verified on the published assets: `readelf -d` on both `Duckle-linux-x64` and `Duckle-linux-arm64` shows no libodbc dependency anymore. So a fresh `sudo apt install libodbc2` should no longer be required.

(The binary still links the usual desktop libraries a Linux GUI app needs, such as GTK and WebKit2GTK, exactly as before; that part is unchanged.)

Please re-download the current v0.5.3 Linux asset (it was replaced in place) and let me know it launches on your machine. Thanks again for the clear report.

LouisDeconinck:
The Linux binaries were re-rolled with unixODBC statically linked and verified to have no libodbc.so.2 dependency. Can this issue be closed?

LouisDeconinck:
@add944 can you close this issue please?

LouisDeconinck:
@SouravRoy-ETL I think you can close this issue.

PR #316, opened 2026-08-31
fix(build): resolve Linux static unixODBC linking and fix build warnings

DarekDan:
## What this changes

Expands native library search paths in `.cargo/config.toml` and `crates/duckdb-engine/build.rs` to support non-Debian Linux distributions when linking static unixODBC (`teradata-static`), and resolves all deprecation and unused code warnings across `duckle-secrets`, `duckle-duckdb-engine`, and `duckle-runner`.

## Related issue

Refs #131

## Type

- [x] Bug fix
- [ ] Performance
- [ ] Docs
- [x] Refactor / internal

## Checklist

- [x] `cargo fmt` and `cargo clippy --workspace --all-targets -- -D warnings` pass

## Notes for reviewers

- Linux distributions outside Debian/Ubuntu (like Arch or Fedora) or environments where static libraries (`libodbc.a`, `libltdl.a`) reside in `~/.local/lib` or `/usr/local/lib` can now build `duckle-runner` with `features = ["teradata-static"]` without manual environment variable workarounds.
- Deprecated APIs (`Nonce::from_slice`, `from_avro_datum`, `Attribute::unescape_value`, and `set_max_row_group_size`) were updated to their modern replacements.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -13,7 +13,20 @@
 # On a non-Debian distro, point ODBC_SYS_STATIC_PATH at your unixODBC lib dir
 # instead (or adjust the path below); on Windows/macOS this file does nothing.
 [target.x86_64-unknown-linux-gnu]
-rustflags = ["-L", "native=/usr/lib/x86_64-linux-gnu"]
+rustflags = [
+    "-L", "native=/home/darekdan/.local/lib",
+    "-L", "native=/usr/local/lib",
+    "-L", "native=/usr/lib/x86_64-linux-gnu",
+    "-L", "native=/usr/lib64",
+    "-L", "native=/usr/lib",
+]
 
 [target.aarch64-unknown-linux-gnu]
-rustflags = ["-L", "native=/usr/lib/aarch64-linux-gnu"]
+rustflags = [
+    "-L", "native=/home/darekdan/.local/lib",
+    "-L", "native=/usr/local/lib",
+    "-L", "native=/usr/lib/aarch64-linux-gnu",
+    "-L", "native=/usr/lib64",
+    "-L", "native=/usr/lib",
+]
+
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -10,23 +10,30 @@
 # targets. The musl target is intentionally absent: the musl runner excludes
 # Teradata entirely.
 #
-# On a non-Debian distro, point ODBC_SYS_STATIC_PATH at your unixODBC lib dir
-# instead (or adjust the path below); on Windows/macOS this file does nothing.
+# Distros that do not use multiarch (Arch, Fedora) keep the archives in
+# /usr/lib64 or /usr/local/lib, so those follow as fallbacks. The multiarch dir
+# stays FIRST: these flags apply to every build for the triple, including the
+# release job that produces the published Linux binaries, and that job must keep
+# linking the same archives it links today.
+#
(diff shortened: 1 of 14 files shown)


SouravRoy-ETL:
Reviewed this across six dimensions (crypto, XML, Parquet, Avro, build/link, and the dead-code suppressions), because green CI does not clear the kinds of change in this diff. Five came back clean, and I want to say so specifically rather than just merging quietly:

- **AES-GCM** (`duckle-secrets`, `runner/build.rs`, `runner/main.rs`): behaviour-preserving. Same 12-byte nonce, same layout, `Cargo.lock` untouched so `aes-gcm` stays at 0.11.1. `try_from` is strictly better than the panicking `from_slice` here, since the length is already guaranteed by the check before `split_at(12)`.
- **`xsd.rs`**: `unescape_value` turns out to be a deprecated alias that forwards to `normalized_value` with identical arguments in quick-xml 0.42, so this is a provable no-op. Worth stating because this repo has a recorded incident where a quick-xml change silently dropped entities, and that is the first thing this diff looked like. It is not that.
- **Parquet**: `set_max_row_group_size` -> `set_max_row_group_row_count(Some(n))` is an exact translation against parquet 59. The old name meant a row count, so the rename is clarifying rather than semantic.
- **Avro**: identity-preserving against apache-avro 0.22.0.
- **`#[allow(dead_code)]`**: each one silences something genuinely unused or test-only. I checked these individually because "the compiler noticed something is not wired up" is a real bug class here, and none of them are that. `is_live` duplicates a check done inline (and a test proves revocation is enforced), `Failure`'s fields are printed before the struct is built, and `handle`/`route_console` are the legacy sync path the axum server replaced.

One thing did need fixing before merge, so I pushed `bc74545` rather than send it back to you:

```toml
[target.x86_64-unknown-linux-gnu]
rustflags = [
    "-L", "native=/home/darekdan/.local/lib",   # <-- and the same under aarch64
```

`rustflags` under a target triple is **not** feature-gated: it applies to every build for that triple, including the release job that produces the published `Duckle-linux-x64` / `Duckle-linux-arm64` binaries. So a committed shared config was naming one person's home directory and putting it ahead of the system lib dirs when choosing which `libodbc.a` to link. On a build host where that path exists, the archive that gets linked changes with no diagnostic either way.

I kept the part you were actually after: `/usr/lib64` and `/usr/local/lib` remain as fallbacks for Arch and Fedora, which is what makes the non-multiarch case work. I moved the multiarch dir back to first so the release builders keep linking exactly what they link today.

Your `build.rs` change already does the per-developer path the right way, and that is the part I would point anyone at: it derives `$HOME/.local/lib` at build time and honours `ODBC_SYS_STATIC_PATH`, gated on `teradata-static`, so it only affects the build that opts in. The config.toml entry looks like local debugging that came along for the ride.

One observation I did **not** change, since it did not survive scrutiny as a defect: `build.rs` emits `$HOME/.local/lib` before `ODBC_SYS_STATIC_PATH`, so an implicit path outranks the one an operator set deliberately. On GitHub runners there is no such directory so nothing turns on it today, but if you touch this again, putting the explicitly configured path first would make the deliberate setting authoritative.

Thanks for the deprecation cleanup, and for `ODBC_SYS_STATIC_PATH` support in `build.rs` - that is the durable fix for #131 on non-Debian distros. Merging once CI is green on the amended branch.

PR #341, opened 2026-09-10
fix(policy): block extension downloads in restricted runs

### LINKED ISSUE #290: Policy: enforce workspace network policy for DuckDB/httpfs external I/O
## Context

#285 added an enforceable workspace/server policy and now covers the engine's own HTTP client at runtime, including dynamic URLs and redirects.

One remaining limitation was explicitly documented in that issue:

> HTTP performed internally by DuckDB through `httpfs` does not pass through Duckle's shared HTTP client and therefore does not pass through those runtime network-policy hooks.

For an AI-agent execution environment, this is an important remaining boundary.

## Problem

A policy may contain:

```yaml
network:
  allowedDomains:
    - api.registry.example
```

and correctly refuse a Duckle `src.rest` request to another domain.

However, agent-authored SQL or a component that causes DuckDB itself to perform external I/O may use a different execution path.

Conceptually:

```sql
SELECT *
FROM read_parquet('https://unapproved.example/data.parquet');
```

or equivalent external access through DuckDB extensions.

If that path bypasses the effective workspace policy, an agent can potentially escape a network restriction without using Duckle's normal HTTP components.

## Desired invariant

In `mode: enforce`:

> A pipeline must not be able to access an external destination that the effective server/workspace policy does not permit, regardless of whether the request originates in Duckle code or inside DuckDB.

## Acceptable implementation approaches

I do not think Duckle needs to intercept arbitrary socket calls inside DuckDB.

A simpler secure implementation is fine.

For example, Duckle could choose one of these approaches:

### A. Enforce approved external destinations

Before executing SQL that performs external I/O, resolve the destination and apply the same effective policy used elsewhere.

or:

### B. Disable unrestricted DuckDB external I/O under enforce mode

Require external data access to go through policy-aware Duckle components / approved connections.

If Duckle cannot prove that an external destination is allowed, refuse it.

This is preferable to a best-effort check that can be bypassed.

## Scope

This should cover at least external access initiated through DuckDB mechanisms such as:

```text
httpfs HTTP/HTTPS reads
S3/object-storage access
other extension-backed external reads/writes where applicable
```

It does not need to become a generic OS sandbox.

Host/container egress restrictions can still be recommended as defense in depth.

## Consistency

The protection should apply regardless of execution surface:

```text
CLI
server/API
scheduler
Plans
MCP
tests
preview
web UI
```

The common execution/compile path should remain the enforcement boundary where possible.

## Tests

Useful RED tests would include:

### HTTP

Policy:

```yaml
network:
  allowedDomains:
    - allowed.example
```

Pipeline SQL attempts to read:

```text
https://blocked.example/data.parquet
```

Expected:

```text
refused before external data is read
```

### S3

Policy allows:

```text
s3://development/
```

SQL attempts to access:

```text
s3://production/data.parquet
```

Expected:

```text
refused
```

### Allowed destination

An explicitly permitted destination should continue to work.

## Why this matters

#285 now provides a strong policy boundary for normal Duckle components.

Closing this remaining execution path would make the statement:

```text
agent cannot bypass network policy by changing how it expresses the same external access
```

substantially stronger.

LouisDeconinck:
Fixes #290

Restricted-network mode already disables DuckDB remote filesystems and community extensions, but explicit `INSTALL` could still download an extension. This change:

- emits `LOAD`-only preludes when the policy denies DuckDB external I/O;
- rejects raw `INSTALL` SQL at the shared DuckDB execution boundary;
- covers stage, probe, batched, and connector-generated extension paths;
- preserves local file access.

Tests:
- `cargo test -p duckle-duckdb-engine install_detector_ignores_text_and_comments`
- `cargo test -p duckle-duckdb-engine a_restricted_network_takes_duckdb_off_the_network_too`
- `cargo test -p duckle-duckdb-engine attach_prelude`
- `cargo test -p duckle-duckdb-engine resource_pragma_tests::no_policy_leaves_duckdb_on_the_network`

### DIFF
--- a/crates/duckdb-engine/src/connectors.rs
+++ b/crates/duckdb-engine/src/connectors.rs
@@ -12250,7 +12250,10 @@ impl DuckdbEngine {
         *ANSWER.get_or_init(|| {
             self.run(
                 None,
-                "INSTALL arrow FROM community; LOAD arrow; SELECT 1;",
+                &format!(
+                    "{}SELECT 1;",
+                    crate::policy::duckdb_extension_prelude("arrow", true)
+                ),
                 true,
             )
             .is_ok()
@@ -16557,7 +16560,8 @@ impl DuckdbEngine {
             .map(|d| format!(", DATA_PATH '{}'", d.replace('\\', "/").replace('\'', "''")))
             .unwrap_or_default();
         let attach = format!(
-            "INSTALL ducklake; LOAD ducklake; ATTACH 'ducklake:{}' AS duckle_src (READ_ONLY{}); ",
+            "{}ATTACH 'ducklake:{}' AS duckle_src (READ_ONLY{}); ",
+            crate::policy::duckdb_extension_prelude("ducklake", false),
             path, data_path
         );
         let node_q = plan::quote_ident(&spec.node_id);
@@ -17260,10 +17264,11 @@ impl DuckdbEngine {
         // Phase 1: stage upstream into a named table that the next CLI
         // invocation will see.
         let stage_sql = format!(
-            "{secret}INSTALL fts; LOAD fts; \
+            "{secret}{fts} \
              DROP TABLE IF EXISTS {staging}; \
              CREATE TABLE {staging} AS SELECT * FROM {upstream};",
             secret = secret_prefix,
+            fts = crate::policy::duckdb_extension_prelude("fts", false),
             staging = staging,
             upstream = upstream,
         );
@@ -17289,12 +17294,13 @@ impl DuckdbEngine {
             None => String::new(),
         };
         let index_sql = format!(
-            "{secret}INSTALL fts; LOAD fts; \
+            "{secret}{fts} \
              PRAGMA create_fts_index('{staging_raw}', '{id_col}', {text_args}); \
              CREATE OR REPLACE TABLE {node} AS \
                SELECT *, {match_expr} AS {output_q} FROM {staging} \
                WHERE {match_expr} IS NOT NULL{order_limit};",
             secret = secret_prefix,
+            fts = crate::policy::duckdb_extension_prelude("fts", false),
             staging_raw = spec.staging_table.replace('\'', "''"),
             id_col = spec.id_col.replace('\'', "''"),
             text_args = text_args,
--- a/crates/duckdb-engine/src/lib.rs
+++ b/crates/duckdb-engine/src/lib.rs
@@ -547,14 +547,7 @@ impl DuckdbEngine {
     /// stdout. Cancellation-aware: polls the child and kills it if a
     /// cancel was requested.
     fn run(&self, db: Option<&Path>, sql: &str, json: bool) -> Result<String, EngineError> {
-        if crate::policy::duckdb_external_io_denied()
-            && crate::policy::contains_explicit_install(sql)
-        {
-            return Err(EngineError::Query(
-                "policy: DuckDB INSTALL is disabled in restricted-network mode; pre-install the extension and use LOAD"
-                    .into(),
-            ));
-        }
+        crate::policy::refuse_install_if_restricted(sql).map_err(EngineError::Query)?;
         if !self.bin.exists() {
             return Err(EngineError::Config(format!(
                 "DuckDB engine isn't installed (expected at {}). Open Setup to install it.",
@@ -3226,6 +3219,15 @@ impl DuckdbEngine {
             }
         }
 
+        // The batched executor is its own CLI entry point: it never calls
+        // `run()`, so the guard there covered the per-stage path and left the
+        // DEFAULT one open. A pure-SQL stage carries its body verbatim, so an
+        // INSTALL in one really did download an extension under an enforcing
+        // policy until this line existed.
+        if let Err(e) = crate::policy::refuse_install_if_restricted(&batched_sql) {
+            return RunResult::failed(total_start, e);
+        }
+
         let mut cmd = std::process::Command::new(&self.bin);
         cmd.arg(db_path);
         // Same as run(): open the throwaway run-db at v1.5.0 so GEOMETRY CRS
@@ -4972,6 +4974,9 @@ pub(crate) fn write_arrayrows_to(
 /// the process env is empty (tests, embedded hosts).
 pub(crate) fn apply_duckdb_sql(bin: &Path, db: &Path, sql: &str) -> Result<(), EngineError> {
     use std::process::Command;
+    // The third CLI entry point, reached from the connectors and the output
+    // cache. Same reason as the other two.
+    crate::policy::refuse_install_if_restricted(sql).map_err(EngineError::Query)?;
     let mut cmd = Command::new(bin);
     #[cfg(windows)]
     {
--- a/crates/duckdb-engine/src/lib.rs
+++ b/crates/duckdb-engine/src/lib.rs
@@ -547,6 +547,14 @@ impl DuckdbEngine {
     /// stdout. Cancellation-aware: polls the child and kills it if a
     /// cancel was requested.
     fn run(&self, db: Option<&Path>, sql: &str, json: bool) -> Result<String, EngineError> {
+        if crate::policy::duckdb_external_io_denied()
+            && crate::policy::contains_explicit_install(sql)
+        {
+            return Err(EngineError::Query(
+                "policy: DuckDB INSTALL is disabled in restricted-network mode; pre-install the extension and use LOAD"
+                    .into(),
+            ));
+        }
         if !self.bin.exists() {
             return Err(EngineError::Config(format!(
                 "DuckDB engine isn't installed (expected at {}). Open Setup to install it.",
@@ -1276,7 +1284,7 @@ impl DuckdbEngine {
             p.push(' ');
         }
         if format == "azureblob" {
-            p.push_str("INSTALL azure; LOAD azure; ");
+            p.push_str(&crate::policy::duckdb_extension_prelude("azure", false));
         }
         // What the RUN path loads for this component, asked OF the run path
         // rather than kept as a second list here.
@@ -7940,7 +7948,7 @@ mod oracle_insert_all_tests {
 
 #[cfg(test)]
 mod resource_pragma_tests {
-    use super::resource_pragmas;
+    use super::{resource_pragmas, DuckdbEngine};
 
     fn guard() -> std::sync::MutexGuard<'static, ()> {
         static L: std::sync::Mutex<()> = std::sync::Mutex::new(());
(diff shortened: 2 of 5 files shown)


SouravRoy-ETL:
Thanks for this - the shape is right, and `duckdb_extension_prelude` routing every emitter through one place is the correct half. The full suite passes on the branch here (2155 passing, 0 failed, with `DUCKLE_DUCKDB_BIN` set to the pinned 1.5.4 CLI).

I can't recommend merging yet though. **The restriction is evadable on both execution paths, and I have runs rather than readings for each.** All three below were reproduced against your branch with the pinned CLI and a `mode: enforce` + `allowedDomains` policy.

## 1. The batched executor never reaches the guard (blocker)

`contains_explicit_install` has exactly one call site, `lib.rs:551`, inside `run()`. `execute_batched` (`lib.rs:2965`) is a second CLI entry point: it appends each stage verbatim at `lib.rs:3092` and spawns its own child at `lib.rs:3229`. It never calls `run()`.

Batched is the **default** path - `target.is_none() && stages >= 2 && all(stage_batchable)`.

```json
{"nodes":[
  {"id":"q1","data":{"label":"pure","alias":"t","componentId":"code.sql",
    "properties":{"pureSql":true,"sql":"INSTALL duckle_no_such_ext_xyz; CREATE OR REPLACE VIEW t AS SELECT 1 AS a;"}}},
  {"id":"k1","data":{"label":"out","componentId":"snk.csv","properties":{"path":"out.csv","hasHeader":true}}}],
 "edges":[{"id":"e1","source":"q1","target":"k1","data":{"connectionType":"main"}}]}
```

```text
$ duckle-runner --pipeline chain.json --workspace .          # batched
error : pure: HTTP Error: Failed to download extension "duckle_no_such_ext_xyz" at URL
        "http://extensions.duckdb.org/v1.5.4/windows_amd64/..." (HTTP 404)

$ duckle-runner --pipeline chain.json --workspace . --target k1   # per-stage
error : pure: query: policy: DuckDB INSTALL is disabled in restricted-network mode
```

Same pipeline, same policy, opposite answers - and the batched one is a live outbound request to `extensions.duckdb.org`, which is the egress #290 is about. (With an extension that exists it simply succeeds.)

`policy.rs:486` says the rejection belongs at "the one boundary shared by stage, probe, and batched execution". `run()` is not shared by batched execution, so that sentence is the thing to fix.

Worth noting this is the repo's recurring shape rather than an oversight of yours - a feature added on one execution path passes the other path's test. Whatever lands here wants a test that runs the **same** pipeline both ways and asserts both refuse.

## 2. A dollar-quoted string desyncs the scanner, on the guarded path too (blocker)

`contains_explicit_install` models `'...'`, `"..."`, `--` and `/* */`, but `$` has no arm and falls to `_ => i += 1`, so the inside of `$$...$$` is scanned as ordinary SQL. An apostrophe in there opens the single-quote branch, finds no close, and consumes to end of input - so everything after it is invisible.

```text
sql: SELECT $$it's$$ AS a; INSTALL duckle_no_such_ext_xyz;
$ duckle-runner --pipeline dollar.json --workspace .     # per-stage, guard active
error : pure: query: HTTP Error: Failed to download extension ... (HTTP 404)
```

It desyncs the other way too: a dollar-quoted body containing the word `install` with balanced apostrophes becomes a false positive.

## 3. The detector refuses legal SQL (major)

`install` is an **unreserved** keyword in the pinned 1.5.4, so it is a legitimate identifier. The check compares every bare identifier token without asking whether it is in statement position:

```text
sql: CREATE OR REPLACE VIEW t AS SELECT 1 AS install;
error : pure: query: policy: DuckDB INSTALL is disabled in restricted-network mode
```

A column named `install` is not far-fetched in this domain, and today it fails only when a policy is active, which is a confusing way to find out.

## A suggestion that removes the lexing problem

Scanning SQL text for a forbidden command has to win every time; an evasion only has to win once, and (1) and (2) are two different ways of losing. DuckDB can enforce this itself:

```text
$ duckdb -c "SET custom_extension_repository='file:///nonexistent';
             INSTALL duckle_no_such_ext_xyz;"
IO Error: Failed to install local extension ... no access to the file at PATH "file:///nonexistent/..."
```

The install is refused **by DuckDB**, with no parsing on our side, so neither a second execution path nor a quoting trick evades it. And it belongs somewhere that already covers both paths: `resource_pragmas` is where `DUCKDB_OFF_NETWORK` is emitted and is applied inside `execute_batched` at `lib.rs:3060`, so one setting there closes both holes at once.

I would keep your lexer as well, for the clear error message - it tells the user to pre-install and `LOAD`, which an IO error does not. But as the message rather than the mechanism.

## One I am less sure about

The diff moves `std::env::remove_var("DUCKLE_POLICY_FILE")` in `a_restricted_network_takes_duckdb_off_the_network_too` from before the assertions to after them (`lib.rs:8012`), while this same PR makes prelude text depend on that process-global. Two reviewers here disagreed: one reproduced cross-test interference under a leaked policy, the other ran the suite and could not make it fail. I could not reproduce a failure either, so treat it as a smell rather than a defect - but the window is real and it would be cheap to clear the variable as soon as the pragmas are captured.

Happy to push these onto the branch if you would rather not round-trip.

SouravRoy-ETL:
One piece of prior art I should have led with, because it argues the same thing from the project's own history rather than from my two repros.

When the DuckDB network boundary was first measured, **scanning SQL was considered and deliberately rejected** as the mechanism. The reasoning recorded at the time: the settings that matter are one-way within a session, so generated SQL cannot lift them, and that is the whole point *because SQL here comes from dbt, Python, templates and MCP* rather than from a person typing it. A denylist over text generated by four different producers has to be right every time.

The two evasions I found are just that argument arriving on schedule.

The same measurement notes recorded `INSTALL` as a **known residual** precisely because `disabled_filesystems` does not cover extension download and `autoinstall_known_extensions=false` does not stop an explicit `INSTALL`. So your read of the gap is right and it was a real hole. `custom_extension_repository` looks like the setting that closes it in the same style as the rest of the boundary, which is why I would put it there rather than in a parser.

SouravRoy-ETL:
**Correction to my suggestion above, before you spend time on it. `custom_extension_repository` is not the boundary I implied, and my "put it in the preamble instead of a parser" advice was wrong.**

I verified it blocks an explicit `INSTALL`. I did not verify the property that actually matters here, and it fails it:

```text
SET custom_extension_repository='file:///duckle-blocked';
SET custom_extension_repository='http://extensions.duckdb.org';   -- accepted
INSTALL something;                                                -- reaches the network again
```

It is **resettable within the session**, so generated SQL can lift it. That is precisely the property the rest of the off-network preamble is built on - `disabled_filesystems` and `allow_community_extensions` are one-way, which is why they hold against SQL arriving from dbt, templates and MCP. A resettable setting is defence in depth, not a boundary.

`SET lock_configuration=true` does make it one-way:

```text
SET custom_extension_repository='file:///duckle-blocked';
SET lock_configuration=true;
SET custom_extension_repository='http://...';
-- Invalid Input Error: Cannot change configuration option ... the configuration has been locked
```

but it locks **everything**, and our own connector preludes set options after the preamble - `SET mysql_enable_transactions` (builders.rs:6181), `SET mssql_insert_batch_size` (:6033), `SET motherduck_token` (:6731). Locking in the shared preamble would break those three sinks in restricted mode, which is a worse bug than the one being fixed.

So **your approach is the right one and I undersold it.** Rejecting the statement is the mechanism; there is no setting that both holds and leaves the rest of the boundary usable. What I said about the parser being fragile stands as a caution, not as an argument for replacing it.

The two blockers are unchanged and both still want fixing: the batched path never reaches the guard, and a `$$...$$` literal desyncs the scanner. The third (bare `install` as an unreserved keyword) is a false positive worth a statement-position check.

Sorry for the detour. I should have checked whether the setting was one-way before recommending it, which is the same standard I was holding the parser to.

SouravRoy-ETL:
Pushed the fixes onto the branch (`ce956a8`) since you had maintainer edits on, rather than sending you round again. Squash it into yours or drop it and redo it your own way, whichever you prefer.

**What changed**

`refuse_install_if_restricted` now holds the refusal, and all three CLI entry points call it: `run()`, `execute_batched`, and `apply_duckdb_sql`. Your comment said the guard belonged at "the one boundary shared by stage, probe, and batched execution" - the honest version is that there is no such boundary today, so the refusal goes to each of them.

The scanner gained a `$` case, tagged and untagged, and a lone `$` stays a positional parameter. And it now looks only where a statement starts, so `SELECT 1 AS install` is ordinary SQL again while `FORCE INSTALL x` is still one command. Comments and whitespace do not move a statement off its opening position, which is what keeps your `/* generated */ INSTALL spatial;` case passing.

**Verified by re-running the four repros, not by reading the diff**

| | before | after |
|---|---|---|
| batched path, `pureSql` INSTALL | downloaded from extensions.duckdb.org | refused |
| `SELECT $$it's$$ AS a; INSTALL ...` | downloaded | refused |
| `SELECT 1 AS install` | refused | runs |
| per-stage INSTALL | refused | refused |

Full suite on the branch: **2157 passing, 0 failed.**

**One thing I deliberately did not add**, because it would have been the smell I raised earlier: a both-paths integration test. Making the policy active needs `DUCKLE_POLICY_FILE`, which is process-global, and asserting on it from a test that runs in parallel with 2000 others is how you get a flake. The new coverage is at the scanner level, and the path wiring is verified by the runs above. If you want a real regression test for it, the honest way is a serialised test that owns the env var, and that is worth doing as its own change rather than smuggled into this one.

Your `duckdb_extension_prelude` half needed nothing - it was already correct on every path, because preludes are built at plan time and both executors inherit them.

SouravRoy-ETL:
Came back to the env-var point I called "a smell", because I had the fix wrong and the actual hazard is sharper than what I described.

I suggested clearing `DUCKLE_POLICY_FILE` "as soon as the pragmas are captured". That does not work here: your two new assertions - the prelude and the `run()` refusal - both need the policy live, so it cannot be cleared before them.

**The real problem is that `assert!` panics.** With the removal after the assertions, the first one that FAILS skips it, and `DUCKLE_POLICY_FILE` then stays set for the remainder of the process. `resource_pragma_tests::guard()` does not save you: it is a module-local mutex, and the other two thousand tests are not holding it. So one failing assertion in this test would have surfaced as unrelated failures scattered across the suite, and the policy now reaches much further than the pragmas it was written for.

That is also why the two reviewers split on it: the mechanism is real but only bites on the failure path, so whether you reproduce it depends on whether something failed.

Pushed `78f0e21`: reads first, clear, then assert on captured values. Nothing about what is asserted changes. I also merged `main` in, so the branch carries the XSD fix and is current.

On `main` separately (`eadcf4c`) I added `DUCKLE_POLICY_FILE` and `DUCKLE_WORKSPACE` to what `guard()` clears on acquisition, so a leak from anywhere is self-healing for this module rather than depending on every test unsetting it on the way out.

Branch suite: **2164 passing, 0 failed.**

PR #342, opened 2026-09-10
feat(xsd): add audited parser-contract acceptance

### LINKED ISSUE #315: XSD: pin accepted schema dependency sets and gate parser-contract drift
## Context

#286 now provides XSD-derived XML parser contracts and records the exact XSD bytes used in run provenance.

Recent work in `4ca1cc6` also resolves multi-file schema sets through:

```text
xs:include
xs:import
xs:redefine
```

and records every resolved schema document as a hashed input artifact in the signed run manifest.

That gives excellent **after-the-fact provenance**.

The remaining gap is **before-the-run enforcement**.

## Problem

A publisher can replace the bytes behind an unchanged path or URL:

```text
https://registry.example/schema/company.xsd
```

while the Duckle pipeline itself remains unchanged.

The next run then derives a different parser contract.

Today the run manifest can tell us afterwards:

```text
previous XSD hash != current XSD hash
```

but the production ingestion may already have parsed and published data using the new contract.

For recurring registry feeds, I would like a mode where changing the authoritative parser contract requires explicit acceptance.

## Desired behavior

Treat the complete resolved XSD dependency set as one versioned parser contract.

Conceptually:

```text
root XSD
    ↓
resolve include/import/redefine closure
    ↓
canonical dependency set
    ↓
hash/fingerprint
    ↓
compare with accepted parser contract
```

A useful policy could be:

```yaml
xsd:
  uri: schemas/company.xsd
  changePolicy: fail
```

with possible values:

```text
allow
warn
fail
```

Backward-compatible default behavior can remain `allow`.

## Dependency-set identity

The fingerprint should cover the resolved schema closure, not only the root document.

For example:

```text
canonical URI/path
SHA-256
resolved namespace, where relevant
```

for every document.

This catches:

```text
root schema changed
import changed
include changed
dependency added
dependency removed
dependency graph changed
```

even when the configured root URI remains identical.

## Fail before parsing the large source

With:

```text
changePolicy: fail
```

the check should happen after the schema set is resolved and hashed but **before** Duckle starts parsing the large XML input or performs downstream side effects.

For a 30 GB registry feed the desired flow is:

```text
resolve XSD set
    ↓
compare contract fingerprint
    ↓
changed?
    yes -> stop
    no  -> start large XML parse
```

A parser-contract mismatch should not advance watermarks, baselines, publication state, or other deferred-success state.

## Explain the change

Machine-readable output should identify what changed.

For example:

```json
{
  "code": "xsd_contract_changed",
  "root": "schemas/company.xsd",
  "previous": "...",
  "current": "...",
  "changes": [
    {
      "uri": "schemas/common-types.xsd",
      "change": "modified",
      "previous_sha256": "...",
      "current_sha256": "..."
    },
    {
      "uri": "schemas/new-types.xsd",
      "change": "added",
      "sha256": "..."
    }
  ]
}
```

Where Duckle can cheaply derive it, also showing the resulting column/schema difference would be useful:

```text
+ legal_form_code VARCHAR
turnover DECIMAL(18,2) -> DECIMAL(20,2)
address required -> optional
```

but dependency-level drift detection is sufficient for v1.

## Explicit acceptance

Provide an explicit way for an operator or authorized agent to make a changed parser contract the new accepted version.

Conceptually:

```bash
duckle schema inspect <pipeline>
duckle schema accept <pipeline>
```

or XSD-specific equivalents.

The exact command naming is not important.

Important semantics:

* acceptance is explicit;
* acceptance is audited;
* a failed run never automatically accepts the observed schema;
* production policy can forbid agents from accepting a new parser contract;
* the previously accepted fingerprint remains available for comparison.

This is similar in spirit to manually accepting a new `qa.baseline`.

## Run provenance

A run should record both:

```text
accepted parser-contract fingerprint
observed/resolved XSD dependency hashes
```

so it is possible to prove that the run used exactly the contract that was approved.

The existing signed artifact records can remain the source of the individual XSD hashes.

## Remote/offline behavior

This should compose with #288 and the network policy.

In a pinned/offline production environment:

```text
accepted XSD dependency set
    -> exact known bytes
```

should be usable without silently fetching a new version from the publisher.

A remote schema changing should never be automatically accepted merely because the URL remained the same.

## Scope

This does **not** require:

* full XSD validation;
* implementing the complete XML Schema standard;
* semantic compatibility analysis between arbitrary XSD versions;
* a hosted schema registry.

A first implementation can simply detect:

```text
accepted dependency fingerprint != observed dependency fingerprint
```

and apply `allow | warn | fail`.

That already provides the important safety property.

## Why this matters

For recurring official-register ingestion, the XSD is effectively part of the parser implementation.

The production invariant should therefore be:

> Pipeline code and parser contract cannot change independently without the change being visible and, when configured, explicitly approved.

That makes registry ingestion considerably more reproducible and safer for unattended or agent-managed pipelines.

LouisDeconinck:
Fixes #315

## Summary
- add `duckle-runner xsd list` to inspect accepted XSD contract fingerprints
- add `duckle-runner xsd accept` requiring an exact SHA-256 and recording a reason
- record actor, timestamp, previous fingerprint, and accepted fingerprint in the existing audit log
- share the contract-store implementation between the connector and CLI
- update the CLI/UI documentation so fail-mode no longer requires hand-editing `.duckle/xsd_contracts`

## Verification
- `cargo test -p duckle-duckdb-engine xsd_contract --lib`
- `cargo check -p duckle-runner`

The runner test binary cannot be linked in this environment because the system ODBC static library is unavailable; compile checking succeeds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2244,8 +2244,14 @@ seen, in `.duckle/xsd_contracts`:
 | `xsdChangePolicy` | on a change |
 |---|---|
 | `warn` (default) | says so once, accepts the new set, run continues |
-| `fail` | refuses the run until you accept it by deleting the line |
+| `fail` | refuses the run until you accept it with `duckle-runner xsd accept` |
 | `allow` | does not look |
+
+Inspect accepted contracts with `duckle-runner xsd list --workspace <dir>`. To
+approve the fingerprint printed by a failed run, use
+`duckle-runner xsd accept --workspace <dir> --uri <schema-uri> --fingerprint <sha256> --reason <why>`.
+The acceptance is written to the audit log with the actor, old and new
+fingerprints, and reason.
 
 The fingerprint covers **every** document in the closure, not the root, because
 an `xs:include` three levels down decides a column's type just as much - a root
--- a/crates/duckdb-engine/src/connectors.rs
+++ b/crates/duckdb-engine/src/connectors.rs
@@ -20743,7 +20743,7 @@ pub(crate) fn xsd_contract_fingerprint(docs: &[(String, String)]) -> String {
 /// behaviour rather than refusing every run.
 pub(crate) fn xsd_contracts_path() -> Option<std::path::PathBuf> {
     let ws = std::env::var("DUCKLE_WORKSPACE").ok().filter(|s| !s.is_empty())?;
-    Some(std::path::Path::new(&ws).join(".duckle").join("xsd_contracts"))
+    Some(crate::xsd_contract::path(std::path::Path::new(&ws)))
 }
 
 /// The contract already accepted for this schema root, if any.
@@ -20752,40 +20752,14 @@ pub(crate) fn xsd_contracts_path() -> Option<std::path::PathBuf> {
 /// can be deleted by hand - which is the whole escape hatch when a publisher
 /// legitimately reissues a schema.
 pub(crate) fn read_xsd_contract(path: &std::path::Path, uri: &str) -> Option<String> {
-    let text = std::fs::read_to_string(path).ok()?;
-    text.lines()
-        .map(str::trim)
-        .filter(|l| !l.is_empty() && !l.starts_with('#'))
-        .find_map(|l| {
-            let (u, fp) = l.split_once(char::is_whitespace)?;
-            (u == uri).then(|| fp.trim().to_string())
-        })
+    crate::xsd_contract::accepted(path, uri)
 }
 
 /// Accept a contract. Best-effort: a workspace that cannot be written still
 /// runs, because failing a run over bookkeeping would be a worse failure than
 /// the one being prevented.
 fn record_xsd_contract(path: &std::path::Path, uri: &str, fingerprint: &str) {
-    if let Some(parent) = path.parent() {
-        let _ = std::fs::create_dir_all(parent);
-    }
-    let existing = std::fs::read_to_string(path).unwrap_or_default();
-    // Replace any line for this uri rather than appending a second one: unlike
-    // a host behind a load balancer, a schema root has exactly one accepted
-    // contract at a time, and two lines would make "which one held?" ambiguous.
-    let mut out: Vec<String> = existing
-        .lines()
-        .filter(|l| {
-            let t = l.trim();
-            if t.is_empty() || t.starts_with('#') {
-                return true;
-            }
-            t.split_once(char::is_whitespace).map(|(u, _)| u != uri).unwrap_or(true)
-        })
-        .map(str::to_string)
-        .collect();
-    out.push(format!("{uri} {fingerprint}"));
-    let _ = std::fs::write(path, out.join("\n") + "\n");
+    let _ = crate::xsd_contract::accept(path, uri, fingerprint);
 }
 
 /// #315: hold the parser contract still, or say plainly that it moved.
--- a/crates/duckdb-engine/src/lib.rs
+++ b/crates/duckdb-engine/src/lib.rs
@@ -84,6 +84,7 @@ pub mod trust;
 pub mod tls;
 pub mod watermark;
 pub mod xsd;
+pub mod xsd_contract;
 mod connectors;
 pub use connectors::remote_fingerprint;
 mod run_log;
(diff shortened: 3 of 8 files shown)


SouravRoy-ETL:
This one looks good to me. Two things I can add that your environment could not.

**The runner tests do link here, and they pass.** Full suite on the branch: **2158 passing, 0 failed** (`cargo test --workspace --exclude duckle-lance --exclude duckle-gpu`, with `DUCKLE_DUCKDB_BIN` pointed at the pinned 1.5.4 CLI). So `xsd_cmd` is covered, not just compile-checked.

**All four generated artifacts are in sync.** I checked rather than assumed, because touching `catalog.json` and `manifest-synth.ts` usually drags the others along:

| artifact | state |
|---|---|
| `crates/duckle-mcp/catalog.json` | in sync |
| `docs/capability-matrix.md` | in sync |
| `website/llms.txt` + `llms-full.txt` | in sync |
| README component counts | unchanged, still correct |

They stay put because the change is a field *description*, and the matrix carries write modes and roles while `llms` carries component summaries. Neither reads field help. I had assumed this PR would red CI on those and was wrong.

The sharing of the contract store between the connector and the CLI is the right call, and requiring the exact SHA-256 rather than a `--force` flag is the detail I like most: it makes "accept this change" mean "accept *this* change".

## Two pre-existing defects this surfaces

Both are on `main` already, so neither is a reason to hold this PR. I am flagging them because `xsd accept` makes the store a documented, user-facing surface, which raises the cost of leaving them.

**1. A schema URI containing whitespace silently disables `fail` mode.** `read_xsd_contract` parses with `split_once(char::is_whitespace)` (`connectors.rs:20760`) and `record_xsd_contract` writes `"{uri} {fingerprint}"` (`:20787`). For `C:/my schemas/a.xsd` the reader recovers `C:/my` as the URI, so the lookup never matches what was written. Every run takes the first-sight branch, and under `xsdChangePolicy: fail` a schema that moved is therefore **never** refused. That is fail-open, and silent - a path with a space is ordinary on Windows.

A length-prefixed or tab-delimited line, or just storing the fingerprint first, would fix it. Whatever changes, the format is now written by two callers, which is an argument for it living in one place - which your PR has already started.

**2. The store is an unlocked, non-atomic read-modify-write.** `read_to_string` -> filter -> `fs::write`, with no lock and no temp-then-rename. `accept` inherits it verbatim from `record_xsd_contract`, so the PR does not introduce it, but it does add a second writer that can now run concurrently with a pipeline - an operator accepting a contract while a scheduled run records one. A torn write leaves a truncated store, and a truncated store reads as "no contract", which is the same fail-open as above.

`runlock::lock_store` already exists for exactly this (it is what `schedules.json` and the alert state use) and would be a small change on both callers.

Happy to open those two as their own issue, or to send a PR for them, whichever you prefer. This one I would merge as it stands once CI has run - it is a fork branch so the workflows need your approval to start.

SouravRoy-ETL:
Fixed the first of the two on `main` in `1610a56`, since it was the fail-open.

**The whitespace one.** Split moved to the last whitespace rather than the first. The fingerprint is a SHA-256 and never contains whitespace, so that is the only unambiguous boundary, and lines written before parse identically because they had no space to be confused by. All three parse sites in `xsd_contract.rs` now go through one function instead of repeating it.

Verified by reverting just that call and watching the two new tests go red, rather than by reading the diff:

```text
a_uri_containing_a_space_is_found_again ......... FAILED
the_previous_fingerprint_survives_a_spaced_uri .. FAILED
```

**The write is atomic now too** - temp file then rename. A bare write truncates first, and a truncated store parses cleanly as "nothing is accepted", which is the same fail-open by another route, and reachable exactly because your PR added a second writer: the engine records a contract at run time while an operator can be accepting one from the CLI.

**What I did not do, and why.** Lost updates between those two writers are still possible. A rename makes the file whole, not the read-modify-write serial - if both read, both edit, both rename, the later one wins and the earlier acceptance is gone. `runlock::lock_store` is the right tool, but `accept` and `accepted` take a store path while `lock_store` wants a workspace, so closing it properly means changing those signatures. That is a real change rather than a line, so it belongs on its own rather than bolted onto a fail-open fix.

I also added a test named for what it actually checks - that the rename leaves no temp behind - rather than calling it an atomicity test, because it is not one: proving atomicity needs a reader racing a writer, and the guarantee there comes from rename being atomic rather than from the test.

Full suite 2161 passing, 0 failed.

PR #346, opened 2026-09-13
test(policy): cover batched extension installs

LouisDeconinck:
Refs #290 — follow-up test coverage for #341.

The merged fix guards every DuckDB CLI entry point (`run()`, the batched executor, and `apply_duckdb_sql`), but the only assertion that a restricted run refuses an `INSTALL` exercises `run()` alone. This adds an end-to-end test proving both executors refuse before a child process is spawned:

- an all-pure-SQL pipeline takes the batched path — the refusal comes from the guard on the batched script;
- a targeted run (`target.is_some()`) takes the per-stage path — the refusal comes from the guard in `run()`.

Both must report `error` naming the disabled INSTALL, so a regression that dropped either guard would fail this test.

Tests:
- `DUCKLE_DUCKDB_BIN=<duckdb> cargo test -p duckle-duckdb-engine --test execution restricted_install_is_refused_in_batched_and_per_stage_runs`
- `cargo test -p duckle-duckdb-engine restricted` (existing restricted-run tests still green)


LouisDeconinck:
The `minio-integration` failure here is unrelated to this diff: MinIO removed their Docker Hub images (`minio/minio` and `minio/mc` both 404 / pull denied since ~Sep 12-13), so the job's `docker pull` fails before anything runs. Every branch hits this — it's not caused by the changes in this PR.

Fix in #347, which switches the job to the official images still published on Quay (`quay.io/minio/minio`, `quay.io/minio/mc`, pinned to release tags).

Everything else is green, and since this PR is a test-only change it isn't affected by the MinIO breakage.

LouisDeconinck:
The `minio-integration` failure was upstream infrastructure, not this PR: MinIO went source-only and removed `minio/minio`/`minio/mc` from Docker Hub, so `docker pull` fails with access denied. Pushed a fix switching the job to the official images still published on `quay.io/minio` — the job should go green on the next run.

LouisDeconinck:
Correction on my earlier note: the MinIO Docker Hub removal fix already exists as a standalone PR — #347 (`fix/minio-quay-images`, pinned quay.io tags). I've reverted the duplicate workflow change off this branch, so #346 stays scoped to the restricted-install fix. `minio-integration` will stay red here until #347 lands and this branch picks it up.

PR #349, opened 2026-09-17
fix(engine): Explode on a non-list column fails with the column type, not length()

### LINKED ISSUE #118: Unnest / Explode node does not work on structs
## What happened

Unnesting astruct throws an error, because `length(STRUCT)` is not defined.

## Steps to reproduce

1. Unnest a column containing a struct

## Environment

- Duckle 0.5.2
- Windows 11 x64

## Logs / error text

```
No function matches the given name and argument types 'length(STRUCT("'v1'" VARCHAR, "'v2'" VARCHAR))'. 
```

LouisDeconinck:
## What this changes

Explode guards NULL/empty inputs with `length(col)`, which DuckDB only binds on list/array columns. On a STRUCT (issue example), MAP, or scalar, the binder rejects the whole stage before the `CASE` runs and the error names `length` - an internal helper - instead of the column or the fix. The stage now runs a one-row `DESCRIBE`-based check first that fails with the column name, its real type, and a pointer to **Flatten** for structs.

## Related issue

Closes #118

## Type

- [x] Bug fix
- [ ] Performance
- [ ] Docs
- [ ] Refactor / internal

## Checklist

- [ ] `cargo fmt` and `cargo clippy --workspace --all-targets -- -D warnings` pass (see notes)
- [x] `cargo test --workspace` passes for the touched crates (`duckle-duckdb-engine` lib + `tests/execution.rs` with `DUCKLE_DUCKDB_BIN`)

## Notes for reviewers

- Why a statement and not an expression: the binder types every `CASE` branch, so `error(...)` inside the SELECT still fails to bind on a struct (verified against DuckDB). A prepended one-row statement that `SELECT error(...)`s only when `DESCRIBE` reports a non-list type is the only formulation that produces a clean diagnostic; it returns zero rows for list columns, so valid pipelines are unaffected.
- The guard also covers **Normalize** with an empty separator (same `length()` guard on an already-array column) but not with a separator, where the column is cast to VARCHAR and any type is fine. Flatten is unchanged.
- The check appears in the exported/Copy-SQL view as a leading statement, which also documents the requirement.
- `cargo fmt --check` is not clean on `main` today (the tree predates default rustfmt), so the new code matches the surrounding style instead. Clippy reports no new warnings in the touched code.


SouravRoy-ETL:
The message this produces is the right message, and a `DESCRIBE` probe is a legitimate way to get it. I checked the three shapes the body names, then went looking for what the probe does to pipelines that work today. Four things, one blocking.

**1. Normalize with no stored `separator` now fails on the column it exists for. This is the blocker.**

`builders.rs:1744` fires the guard when `string_prop(props, "separator").unwrap_or_default().is_empty()`, and `string_prop` returns `None` for an absent key (`builders.rs:9978`). The builder that produces the body underneath defaults an absent separator to a comma instead: `builders.rs:2037`, `unwrap_or_else(|| ",".into())`. So the two halves disagree about what "no separator" means.

That matters because the form only *displays* the manifest default. `PropertiesPanel.tsx:651-655` passes `props[key] !== undefined ? props[key] : field.defaultValue` and writes a prop only in `onChange`, so a Normalize node dropped on the canvas, given a column, and otherwise left alone has no `separator` key at all. After this change that node dies with `Invalid Input Error: Normalize needs a list/array column; "tags" is VARCHAR`, although the stage would have split on the comma correctly. Same for any hand-authored pipeline JSON that omits the key. The new unit test asserts the current behaviour for `{"column":"tags"}`, so it pins the mismatch rather than catching it.

Fix: fire only when the prop is present and empty, `matches!(string_prop(props, "separator").as_deref(), Some(""))`, and flip that assertion to expect `None` for `{"column":"tags"}`.

**2. A fixed-size ARRAY is refused, by a message that contradicts itself.**

`builders.rs:1763` tests `column_type NOT LIKE '%[]'`. DESCRIBE renders a LIST as `INTEGER[]` but a fixed-size ARRAY as `INTEGER[3]`, identically on 1.2.2, 1.4.4 and 1.5.4, so this is not version drift. A `code.sql` stage producing `array_value(1,2,3)`, or any `CAST(x AS FLOAT[384])` vector column, explodes correctly today and afterwards fails with `"v" is INTEGER[3]` - telling the author that an array column is not an array column. Only LIST and ARRAY end in `]`, so `NOT LIKE '%]'` covers both. Worth a `[1,2,3]::INTEGER[3]` case in the regression test.

**3. Prepending a statement to `Stage.sql` turns off node analysis for every Explode node.**

`analyze_node_sql` refuses any stage whose SQL does not start with `CREATE OR REPLACE VIEW` (`lib.rs:3942-3946`), and it runs on node selection from the desktop, from `serve` and from the properties panel. With the probe first, selecting an Explode node returns `validated: false`, no columns and no diagnostics; `schema-resolve.ts:125-131` then clears any problem previously reported for that node and keeps the static guess. A mistyped column goes silently unflagged in the panel, and the exploded column keeps its LIST type downstream. Keeping the CREATE first - carry the probe in its own field that the executor prepends - fixes this and 4.

**4. Column lineage for an Explode stage degrades to passthrough.** `select_body` (`lineage.rs:220`) now sees a multi-statement string, `json_serialize_sql` refuses it, and `lib.rs:1272` swallows the error with `unwrap_or_default()`, so the node is treated as transparent instead of resolving its own columns. Mild, because an explode's output names match its input's, which is also why nothing would notice.

**On the body, for the record.** `length` does bind on VARCHAR, so a VARCHAR scalar fails later with `Cannot mix values of type VARCHAR and "NULL"[] in CASE expression` and never mentions `length`; the STRUCT, MAP and INTEGER halves reproduce exactly as you describe. And this does not close #118 the way the issue records it: the plan there is to extend the existing diagnostics layer (`sqldiag.rs`, already used by `analyze_node_sql`) to generated SQL, which costs no extra statement per run and would have none of 1 to 4. I am not against the probe, but if it stays, it has to be the version that does not regress working pipelines.

**Cost, since it is easy to miss.** The probe is unconditional: one extra statement per Explode or empty-separator Normalize stage, per run, on both execution paths, never gated on failure. It is bind-only, measured around 4 ms over a view on an 8 MB CSV source, but it re-binds the upstream, so a lazy view over a remote file pays a second metadata fetch. It is also skipped entirely for `materialize` disk/duckdb nodes, where the executor ignores `Stage.sql`.

**Tests.** Both new execution tests are genuinely red without the change. The gaps are the four shapes above, plus the per-stage executor: all three new pipelines are batchable, so they exercise the batched path only, and the standing trap in this engine is that a change wired into one path passes the other path's test.

LouisDeconinck:
All addressed in the latest push:

- The guard now only fires when `separator` is present and empty (`matches!(..., Some(""))`) - an absent separator keeps the body's comma default and never reaches the guard.
- The column-type probe accepts any `]`-suffixed type, so fixed-size `INTEGER[3]` arrays explode instead of erroring.
- The probe no longer prepends into `Stage.sql` (which broke `analyze_node_sql` and lineage). It travels on a new `Stage.pre_sql` field; the executor runs it before the CREATE on both the batched and per-stage paths.
- Regression tests cover: absent separator, explicit empty, explicit non-empty, `ARRAY[N]`, the STRUCT diagnostic, and both execution paths.

SouravRoy-ETL:
Verified all four on 73531b4, and the interesting one is the third.

The separator guard now reads `matches!(string_prop(props, "separator").as_deref(), Some(""))` (`builders.rs:1747`), which agrees with `build_normalize`'s own default for every JSON shape: absent, null and a non-string all fall through to the comma, and only an explicit empty string means array mode. The flipped unit test at `plan/tests.rs:2604-2608` pins it, with the explicit-empty and explicit-comma cases either side. `NOT LIKE '%]'` admits fixed-size arrays without opening a hole, since STRUCT, MAP, DECIMAL and UNION all render with a trailing paren, and `explode_on_a_fixed_size_array_column_still_explodes` covers it.

Moving the probe into `Stage.pre_sql` rather than stripping it before the prefix check is the better of the two options I offered. `stage.sql` is CREATE-first again so `analyze_node_sql` binds it, both executors prepend the field (`lib.rs:1950-1956` per-stage, `lib.rs:3171-3179` batched, after any BEGIN), and the probe correctly does not run during analysis, where it would only be noise against zero-row stubs. That also fixes the lineage passthrough for free, since `select_body` is handed one statement again. I checked both `Stage` literals in the repo set the new field.

Two small things:

The SQL export no longer shows a statement the run executes. `compile_pipeline_sql_opts` builds each `StageSql` from `s.sql` alone (`lib.rs:7163-7177`), and `StageSql` has no `pre_sql`, so a copied-out script omits the guard. On the reviewed head it was included, because it lived inside `s.sql`. That matters less than it sounds, since a passing probe is a no-op and the data is identical, but the same function re-synthesises a publish group's BEGIN/COMMIT a few lines below precisely so the exported script is the script that runs, and it is one `unwrap_or("")` to keep that true. It also changes the error a user gets when they paste the script: ours names the column and points at Flatten, the exported one says `No function matches length(STRUCT(...))`.

`explode_prepends_a_type_guard_that_names_the_column` (`plan/tests.rs:2575`) now asserts the guard is NOT prepended. Worth a rename so the next reader does not put the prepend back.

One coverage gap I could not settle without a run, so treat it as a question rather than a finding: nothing puts an Explode directly downstream of an extension-ATTACH source. The batched executor deliberately avoids `COUNT(*)` and the DESCRIBE preview over such a relation (`lib.rs:3230-3247`) because of a DuckDB binder bug, and the probe issues `DESCRIBE SELECT * FROM <upstream>` with no such exclusion. I expect it is fine, since the downstream node's own preview already binds through that relation. If you have a Postgres source handy it is a two-minute check.

LouisDeconinck:
Both done in db2adee: `compile_pipeline_sql_opts` now prepends `s.pre_sql` to each exported `StageSql` (same terminator handling as the batched executor), ahead of the CREATE and inside a publish group's BEGIN/COMMIT — a copied-out script is again the script that runs. New `export_includes_a_stage_pre_statement` pins it. The regression test is renamed to `explode_carries_a_type_guard_that_names_the_column`.

On the extension-`ATTACH` question — verified against a live Postgres in one `duckdb -bail` session on 1.5.5: `ATTACH ... TYPE postgres` + `CREATE VIEW s AS SELECT * FROM duckle_src.public.t`, then the probe's `FROM (DESCRIBE SELECT * FROM s)` binds cleanly — zero rows, explode produces the right rows. A VARCHAR column gives `Invalid Input Error: Explode needs a list/array column; "name" is VARCHAR` as intended. No exclusion needed: the binder bug only hit the `literal`-plus-`COUNT(*)` projection, not a `DESCRIBE`/`SELECT *` bind.

SouravRoy-ETL:
Both items are done as asked: the export prepends `s.pre_sql` with the executor's own terminator handling and ahead of the BEGIN/COMMIT synthesis (`lib.rs:7184-7192`), `export_includes_a_stage_pre_statement` pins it, and the test is renamed to match what it asserts. Your answer on the extension-ATTACH question is good enough for me, and the `duckdb -bail` run against Postgres settles it better than my reading did.

The export change has a consequence neither of us saw, though, and it puts round two's third finding back.

**`analyze_node_sql` reads the export, not the plan.**

`lib.rs:3931` is `let stages = compile_pipeline_sql(doc)?` - the function this commit just changed. So the `StageSql` it inspects now begins with `SELECT error('Explode needs a list/array column; ...') ... LIMIT 1;`, and the gate at `lib.rs:3955-3960` requires `stage.sql` to start with `CREATE OR REPLACE VIEW`. It does not, so the node comes back `validated: false` with the note "does not compile to a plain view", no columns and no diagnostics.

That is the same end state as before: `schema-resolve.ts:124` clears any problem previously reported for the node, and with no columns the panel keeps its static guess, so a mistyped column on an Explode node reads as clean. `describe_node_columns` is worse - `lib.rs:3744` turns that note into an `Err`, so the desktop command and the serve route now fail on a node that works on `main`.

Column lineage really is fixed, and the contrast is the tell: `pipeline_column_lineage` takes `plan::compile(doc)` (`lib.rs:1253`) and reads `Stage.sql` directly, so it never sees the probe. Analysis is the one consumer that goes through the export.

Two ways out, both clean:

- point `analyze_node_sql` at `plan::compile(doc)` the way lineage already does, so it sees `Stage.sql` and `StageKind` rather than rendered text, or
- give `StageSql` its own `pre_sql: Option<String>` and let the Plan tab and the export renderer join the two, leaving `StageSql.sql` a single statement.

The second keeps the export honest without handing a two-statement string to anything that binds it, which is probably what you want long term.

Either way this needs the test neither head has: `analyze_node_sql` on an `xf.arr.explode` node returns `validated: true` with columns, and reports a diagnostic for a mistyped column. All thirteen checks are green on db2adee because every analyze/describe test in `tests/execution.rs` uses a non-Explode node.

**One nit while you are in there**: `sql` goes through `redact_secret_values` at `lib.rs:7176`, but `pre` is interpolated raw at `:7185`. Nothing leaks today, since the only producer is the type guard, but the next component to carry a pre-statement is most likely an ATTACH prelude, and that is where credentials live. `redact_secret_values(pre, &secrets)` is a one-call fix.

Issue #312, opened 2026-08-30
CI: add stable versioned JSON plus JUnit/SARIF output for validate, test, contract, policy, and review


LouisDeconinck:
## Problem

Duckle has strong CLI validation, tests, review, contracts, and policy findings.

CI systems and coding agents benefit when results use standard formats rather than requiring custom parsing of console text.

## Proposed solution

For relevant commands, support:

```text
--format json
--format junit
--format sarif
```

Commands:

```text
duckle validate
duckle test
duckle contracts check
duckle policy check
duckle review
```

## Stable JSON

Version the machine-readable schema:

```json
{
  "schema_version": 1,
  "command": "validate",
  "findings": []
}
```

Document stable exit codes for:

```text
success
validation/test failure
configuration/tool failure
policy refusal
```

## JUnit

Represent tests/checks as test cases with:

```text
pipeline
node/assertion
duration
failure message
captured safe diagnostics
```

## SARIF

Use SARIF for file/node-level findings such as:

```text
unknown property
invalid SQL
contract break
policy violation
unsafe component
schema drift
```

Map to pipeline/workspace file and node/property where possible.

## GitHub annotations

SARIF is sufficient for GitHub Code Scanning; optional concise workflow annotations can be generated without embedding GitHub-specific behavior into the core engine.

## Security

Generated reports must use the same secret/sensitive-value redaction as logs and API results.

## Acceptance criteria

1. CI can display failed Duckle tests as native test results;
2. policy/validation findings point to the relevant file/node/property;
3. JSON schemas and exit codes are documented/versioned;
4. reports never expose secrets;
5. output remains consistent across platforms.

SouravRoy-ETL:
Shipped for `validate` on `main` (`d8958a2`): `--format json | junit | sarif`.

**Exit codes are documented and distinct**, which I think matters more than the formats:

| code | meaning |
| --- | --- |
| 0 | everything checked passed |
| 1 | a check failed - the thing being checked is wrong |
| 2 | the tool could not run - bad flag, unreadable file, no input |

1 and 2 are deliberately separate. A job usually wants to fail differently on 2, because 2 means the gate never actually ran, and treating that as a pass is how a broken gate goes unnoticed for a month.

[...]

PR #350, opened 2026-09-17
feat(runner): review --format json|junit|sarif matches the other gates

LouisDeconinck:
## What this changes

Adds `--format json|junit|sarif` to `duckle-runner review`, emitting through the same `report` module that `validate`, `test` and `contracts check` already use, so a CI job reads one format across every gate. Compile/run/drift failures gate; node diffs and row-count changes surface as informational passes. `--json` and the exit codes are unchanged.

## Related issue

Refs #312 - `test` and `contracts check` already had `--format`; this completes the review side. There is no `policy` subcommand today, so nothing was needed there.

## Type

- [ ] Bug fix
- [x] New feature (connector / transform / sink)
- [ ] Performance
- [ ] Docs
- [ ] Refactor / internal

## Checklist

- [ ] `cargo fmt` and `cargo clippy --workspace --all-targets -- -D warnings` pass (see notes)
- [x] `cargo test --workspace` compiles the runner; `review --format` smoke-tested for json, junit, sarif, bad-format and exit codes

## Notes for reviewers

- `--format json` wraps the existing `--json` document in the versioned findings envelope (schemaVersion/command/ok/checked/failed/findings), so the detail is still there for anyone who parsed it.
- JUnit lists one testcase per compile check and per node diff; SARIF emits only failures (the compile/drift/run gates).
- `cargo fmt --check` is not clean on `main` today, so the new code matches surrounding style instead. Clippy reports no new warnings in the touched code.


SouravRoy-ETL:
The wiring is right: this goes through the same `report` module as `validate`, `test` and `contracts check`, the `--format` arm matches their shape and wording, a bad value returns `Err` so the dispatcher exits 2, and `--format` takes precedence over `--json` exactly as in `validate`. One thing to change before it goes in.

**A before-side failure is reported as a failing check while the gate exits 0.**

Review a change that *fixes* a broken pipeline: `--before old.json --after new.json --format junit`, where the old version does not compile and the new one does. `main.rs:1950-1955` loops over both sides and pushes `Finding::fail(file, "compile", ...)` for either, so the JUnit document carries `failures="1"` with a `<failure>` element and the JSON envelope reports `"ok": false`. The gate at `main.rs:2159` inspects only `after_compiles`, `after_run_failed` and the after-side drift, so the process exits 0. A CI job that renders that file shows a red test for a PR that repaired the pipeline, and `--format sarif` puts an `error` annotation on the old file's path, which is usually a temp extract of the previous revision rather than a file in the repository.

The same inversion happens with `--data` when the before version no longer runs, which is the ordinary case when a change moves a source file: `main.rs:1977-1983` emits a `data-run` failure, `after_run_failed` at `main.rs:1883` is false, exit 0.

This is also a contract mismatch rather than only a cosmetic one. `report.rs:19` documents exit 0 as "everything checked passed", `main.rs:1064` documents review's own 1 as "the `--after` version fails to compile", and every sibling keeps the two in step (`validate` exits 1 iff a finding failed, `main.rs:1739`; `test` likewise). `contracts check` is the one precedent for `ok: false` with exit 0, and I would treat that as an inconsistency there rather than a licence here.

Either of these settles it:
- report the before side as information, e.g. `Finding::pass(file, "compile-before", format!("before does not compile: {e}"))`, which matches the documented exit semantics and needs no other change, or
- add before-side failures to the gate at `main.rs:2159`, which changes the exit-code contract and so needs the help text and the README changed with it.

I would take the first.

**Two smaller notes.**

The claim that compile, run and drift failures gate is half right: node diffs and row counts are indeed informational passes (`main.rs:1963-1970`, `1996-2005`), but only the after side gates, which is what the finding above is about.

README sync: `README.md:1881-1890` is the section that documents this exact family ("Reports CI already understands (`--format`)") and still lists only `validate` and `test`, and `README.md:1920` states a family rule that is not true for review. A surface change here normally lands with its README line in the same commit.

**Tests.** None are added and none would be red without the change, which is why the exit-code inversion above is invisible to the suite. `main.rs` has no test module at all, so I am not asking you to add the first one lightly - but the cheap version is to extract the findings construction into a free function over the before/after results and assert two things: a before-only compile failure produces no failing finding, and a changed node's message names the node. `report.rs:219-307` already covers the emitters themselves, so that is the only uncovered piece.

For the checklist: CI is green here, but note that the fmt and clippy steps are `continue-on-error: true` in `ci.yml:194-199`, so a green run does not attest either of those two boxes.

LouisDeconinck:
Done:

- Before-side compile and data-run failures are now informational passes (`compile-before` / `data-run-before` findings) - a broken old version is the expected case when reviewing a fix, and only the after side gates.
- Findings construction is extracted into a testable `review_findings` helper; new tests cover a before-only compile failure producing no failing finding, and changed-node findings retaining the node id.
- README documents `review --format json|junit|sarif` and that the exit gate is the after version.

SouravRoy-ETL:
Both fixed, and I checked the exit code against the report rather than only the reported case: `ok` and the exit status now agree on all four failing-finding paths, not just the before-side compile one. The before side is informational (`compile-before` / `data-run-before`), the gate at the bottom is untouched, and the README family section lists review alongside validate and test.

One nit, no action needed unless it bothers you: the before-side rule id changes with the outcome, so a CI job that groups by rule sees `compile-before` and `compile` as two different checks for the same stage, and the error text rides in the message of a passing finding, which JUnit and SARIF render less prominently than a failure. That is the correct trade for not gating on it. Just noting it is a trade.

Ready from my side.

Issue #90, opened 2026-06-22
Make more samples available


stepro71:
Great project!

Would it be possible to add sample projects to source code like those displayed in https://duckle.org/use-cases.html

SouravRoy-ETL:
I will check the options to enable it.

vrejany:
Do subscribe to this request, especially for the flow control nodes.

stepro71:
done

LouisDeconinck:
One way to make this finite and maintainable would be to treat the samples as **runnable regression-grade example workspaces**, rather than screenshots or static documentation.

A small initial set could cover:

- REST/API → transform → database/file sink
- incremental/file ingestion with schema changes
- join/lookup + mapping
- flow control: branch / wait / failure handling
- contexts/environments + secret references
- a `duckle test` example with fixtures and assertions

If the examples run in CI, they serve three purposes at once: tutorials, reference implementations, and regression tests.

I would prioritize the flow-control example requested above because those components are much harder to understand from the palette alone than normal SQL transforms.

PR #351, opened 2026-09-17
feat(examples): runnable flow-control workspace, exercised in CI

LouisDeconinck:
## What this changes

Adds `examples/flow-control`, a runnable workspace demonstrating the `ctl.*` components the issue discussion specifically asked for (they are hard to understand from the palette alone):

- **route_orders** - `ctl.switch` branch routing, `ctl.die` as a reject-branch guard, `ctl.log`/`ctl.warn` run-log messages
- **foreach_region** - `ctl.wait`, `xf.distinct`, `ctl.foreach` with `itemKey` and `${ITER_ITEM_*}` substitution in the child pipeline
- **try_fallback** - `ctl.try` with a `fallbackPipelineRef` child

Data is committed (`data/orders.csv`), so every pipeline runs with zero setup and no network.

## CI

New steps in `ci.yml`: every example pipeline is `validate`d on all platforms, and on the Linux leg (runner + `DUCKLE_DUCKDB_BIN` already there) each top-level pipeline is run for real - starter-workspace's `gen_samples.sql` is generated first, `_`-prefixed children run through their parents.

## Related issue

Refs #90

## Type

- [ ] Bug fix
- [x] New feature (connector / transform / sink) - sample/workspace content
- [ ] Performance
- [ ] Docs
- [ ] Refactor / internal

## Checklist

- [ ] `cargo fmt` and `cargo clippy --workspace --all-targets -- -D warnings` pass (no Rust changes)
- [x] `cargo test --workspace` passes (no code changes)

## Notes for reviewers

- Verified locally: all three new pipelines run to `status: ok` and produce the expected `output/` files; all six starter pipelines also run after `gen_samples.sql`.
- `pipelineRef` values use the full `.pipeline.json` filename because `resolve_subpipeline_ref` matches `<stem>.json` or a literal path - a bare id like the starter's never resolves.
- The child-prefixed-`_` convention is carried over from starter-workspace.
- If a different directory layout is preferred (e.g. one workspace per use-case page), happy to move things around.


SouravRoy-ETL:
The workspace itself is what I wanted from this: `ctl.*` is the part of the palette nobody can learn from tile names, the data is committed so it runs with no setup, and putting the gate inside the existing `rust` matrix job rather than a standalone one is exactly the lesson `ci.yml:159-163` records about system dependencies. Both new steps already passed on the Linux leg, so the CI work is proven.

One change needed before merge, and it is the one-line part.

**The csv_split `pipelineRef` change breaks that sample in the workspace the file actually ships to.**

`examples/starter-workspace` is not only run from the checkout: `apps/desktop/src/samples.rs:15` embeds it with `include_dir!`, and seeding strips the suffix and writes each pipeline as `<id>.json` (`samples.rs:53-56`). The crate's own test pins both halves - `pipelines/_csv_split_child.json` exists, and the `.pipeline.json` name does not (`samples.rs:149,167`).

So after this change, a new user who opens an empty folder in the desktop app and runs sample 6 gets a failure. `resolveForRun` rewrites a `pipelineRef` only when it equals a repository item's id or name (`run-resolve.ts:378-382`, `411-415`); `_csv_split_child.pipeline.json` is neither, so it reaches the engine unchanged, and none of the resolver's candidates exist in the seeded layout. Before this change the bare id matched the repository id, was rewritten to the real path, and the seeded sample worked. The bug you hit is real, but it is specific to running the checkout in place.

I have fixed the underlying cause on `main` in `440a889`: `resolve_subpipeline_in` now strips a trailing `.pipeline` from the stem and looks for both `<stem>.json` and `<stem>.pipeline.json`, in the `pipelines/` candidate and in the workspace walk, with a test that resolves the same bare reference in both layouts. So please revert this line to `"pipelineRef": "_csv_split_child"` and rebase - the headless run your new CI step performs will then resolve, and the seeded sample keeps working. That also de-risks the new flow-control references, which today only work because that workspace is never renamed.

**Two things in the body to correct for the record.** The bare id did resolve in the desktop and web run path, via the id-to-path map above, so "never resolved" holds only for a file-loaded headless run. And the `_`-prefixed children do not all run through their parents: `_region_export` and `_csv_split_child` do, but `_notify_error` never executes, because `try_fallback` succeeds and `ctl.try`'s fallback only fires on a later-stage failure. It is compile-checked by `validate` and nothing more, which is worth saying in the README for the workspace so a reader does not assume the fallback path is exercised.

**The cancelled Linux leg is not yours.** The `rust (ubuntu-24.04)` job hit the six-hour timeout inside `cargo test`: the last line is `rabbit_source_and_sink_open_a_socket_to_the_broker has been running for over 60 seconds`, then nothing until the runner killed the orphan test process. That test opens an AMQP socket to a local fake broker and this PR touches no Rust. I will deal with that test separately; re-running the leg is all this needs.

**Two suggestions, neither blocking.**

`for ws in examples/*/` with `"$ws"pipelines/*.pipeline.json` passes the literal glob to `validate` if a future example has no `pipelines/` directory, which would fail the step on all three runners. `[ -d "$ws/pipelines" ] || continue` settles it.

The run step checks exit status and row counts but never that the outputs exist, so a sink that reports rows and writes nowhere would pass. One `test -s examples/flow-control/output/by_region/North.csv` on the Linux leg would close that.

And one line in `README.md:3856`, where "Recipes and examples" currently points only at `samples/`: `examples/` is mentioned nowhere in the README or docs today, so both workspaces are undiscoverable except by browsing the tree.

SouravRoy-ETL:
Follow-up on the cancelled Linux leg, since it was in the way of reading this PR rather than anything you did.

The cause is fixed on `main` in `c7224ff`. `lapin::Connection::connect` carries no timeout of its own, so the fake broker in that test - a plain TCP listener that accepts and never sends the AMQP Start frame - left the await pending indefinitely, and the node's `timeoutMs` could not end it because that deadline is only evaluated once the connect has returned. Both `src.rabbit` and `snk.rabbit` now connect through one helper with a 30 second budget, the same one the HTTP transports use, and report `connect: the broker accepted the connection but did not answer within 30000ms` instead of waiting.

I have re-run the leg. Nothing needed from you for it.

LouisDeconinck:
Done:

- Rebased on main; `pipelineRef` is back to the bare `_csv_split_child` form now that the resolver fix is in (`440a889`).
- The CI example loop guards with `[ -d "$ws/pipelines" ]` and now asserts an expected output file exists (`examples/flow-control/output/by_region/North.csv`), so a silently empty run fails.
- README mentions the runnable `examples/` workspaces.

SouravRoy-ETL:
Verified on b532f6f: the bare `_csv_split_child` is back, `440a889` is in the ancestry so it resolves in both layouts, the workspace README says `_notify_error` is compile-checked rather than run, the run step asserts the outputs exist, and the README points at `examples/`. Green on every leg.

Three small things:

The `-d pipelines` guard went into the validate loop but not the run loop (`ci.yml:197`), so a future example without a `pipelines/` directory would still expand a literal glob there and fail the step on all three runners. Same one-line guard.

The new README sentence describes both workspaces as having committed data, but starter-workspace's data is generated by its own script and gitignored. Only flow-control's `orders.csv` is committed. Worth a word so a reader does not clone it and expect rows.

And the PR body still says the bare id "never resolved", with the full `.pipeline.json` filename as the rationale. That is the claim this change now contradicts: it resolved fine in the desktop and web run path through the repo-item map, and the full filename is what broke the seeded layout. Not worth a force-push on its own, but if you touch the branch again it would be good not to leave that in the history.

Ready from my side.

SouravRoy-ETL:
Both done in 5fa5618: the run loop has the same `-d "$ws/pipelines"` guard as the validate loop, and the README sentence now distinguishes flow-control's committed `orders.csv` from starter-workspace's generated, gitignored data. All thirteen checks green on that head.

Nothing left from me except the PR body, which still says the bare id "never resolved" and gives the full `.pipeline.json` filename as the reason. The change itself contradicts that now, and the resolver fix on main is what makes the bare ref work in both layouts. Worth editing the description before this goes in, so the history does not record a rationale that was wrong.

Ready from my side.

Issue #308, opened 2026-08-30
Git/catalog: calculate and optionally run changed or transitively affected pipelines


LouisDeconinck:
## Current state

Duckle already has:

- Git integration and revision diffs;
- a cross-pipeline asset graph;
- downstream impact traversal;
- validation/tests;
- Plans and durable runs.

These primitives can support a simple affected-pipeline workflow.

## Problem

A change to a shared producer, SQL routine, contract, or pipeline may affect many downstream pipelines.

CI and operators need to answer:

```text
What changed?
Which pipelines are directly changed?
Which downstream pipelines may be affected?
What should be validated/tested/run?
Why is each pipeline included?
```

## Proposed solution

Add commands such as:

```bash
duckle affected --base main --head HEAD --json
duckle validate --affected --base main
duckle test --affected --base main
duckle run --affected --base <active-release>
```

## Change inputs

Consider changes to:

```text
pipeline definitions
Plans
shared routines/templates
component/runtime lock
contracts/schema declarations
contexts that affect compilation
external component definitions
```

Secret value changes should not be read from Git.

## Impact

Use the catalog graph to include transitive downstream consumers.

For unresolved/dynamic dependencies, return `uncertain` and include conservatively when requested.

## Explanation

For every result, expose a reason chain:

```text
pipeline normalize_company changed directly

pipeline serving_company affected because:
normalize_company -> canonical_company -> serving_company
```

## Execution

When running affected pipelines:

- dry-run by default;
- topologically order dependencies where possible;
- run each pipeline once;
- use durable runs/resource pools;
- bind all runs to the same candidate release;
- stop or continue according to an explicit failure policy.

## Scope

This does not need an always-on auto-materialization daemon.

A deterministic CLI/API operation for CI and release promotion is enough.

## Acceptance criteria

1. a changed producer identifies transitive downstream pipelines;
2. every included pipeline has an explanation;
3. unresolved dependencies are visible, not silently ignored;
4. CI can consume versioned JSON;
5. optional execution uses normal run/pool semantics.

SouravRoy-ETL:
Shipped on `main` (`81d47fe`).

```bash
duckle-runner affected --base main                  # against the working tree
duckle-runner affected --base main --head HEAD --json
duckle-runner validate --affected --base main       # validate only what it reaches
```

```text
affected against main (head: working tree)

  produce                      changed
  middle                       produce -> lake/orders.parquet -> middle
[...]

Issue #204, opened 2026-07-25
All context variables are merged into a single flat map where later contexts overwrite earlier ones with the same key. This causes silent data loss when multiple context items define the same variable name, with no visibility or warning to the user.


jaty613:
frontend/src/run-resolve.ts:33-46 的 buildContextVars：

export function buildContextVars(repo: RepoItem[]): Record<string, string> {
    for (const item of repo) {
        if (item.type !== 'context') continue;
        for (const v of payload.variables) {
            out[v.key] = v.value;  // overwrites with the same key directly
        }
    }
    return out;
}

Problem: All context variables are merged into a single flat map where later contexts overwrite earlier ones with the same key. This causes silent data loss when multiple context items define the same variable name, with no visibility or warning to the user.

Proposed solutions:

Conflict detection — Throw an error when duplicate variable keys are detected across contexts.

Namespace scoping — Prefix variables with context IDs or names to avoid collisions.

Warning logging — At minimum, log warnings when overwrites occur.

SouravRoy-ETL:
Thanks for flagging this. v0.5.8 addresses it: when two context items define the same variable name, validation now raises a warning for the colliding key instead of silently overwriting one with the other, so the shadowed value is visible rather than causing quiet data loss.

Please grab the v0.5.8 build and let us know if it behaves as you expect:
https://github.com/slothflowlabs/duckle/releases/tag/v0.5.8

jaty613:
Thanks for the warning. That prevents silent data loss, but it doesn’t address the underlying environment management need.
Right now the only supported pattern is “one active context at a time”. If I want a common base + env-specific overrides (or to compose multiple contexts), the current flat-merge + warning still forces me to manually resolve collisions every time.
Would you consider supporting context inheritance / layering (or at least an option to treat collisions as hard errors instead of warnings)?

SouravRoy-ETL:
Yes, please expect it in the next release. Stay tuned.

LouisDeconinck:
Context layering has now shipped, including deterministic override precedence and collision detection. That seems to address the remaining request here. Can this issue be closed?

LouisDeconinck:
@jaty613 can you close this issue please?

LouisDeconinck:
@SouravRoy-ETL I think you can close this issue.

PR #352, opened 2026-09-17
feat(runner): test --affected runs only the suites a change reaches

LouisDeconinck:
## Summary
- `duckle-runner test --affected --base <rev>` keeps only the test suites whose pipeline the base..head diff reaches, using the same `affected` selection `affected` prints and `validate --affected` already gates on - the three consumers can no longer disagree about what changed
- Skipped suites print by name; a selection that resolves to no suites exits 0 with `nothing affected against <base>` so a clean pull request does not fail its own gate, and a selected pipeline that cannot be located errors instead of silently passing
- New `tests/affected_test.rs` exercises the real binary against a real git workspace: clean tree passes, an edited pipeline runs its suite and skips the neighbour's

## Related issue
Refs #308 - `affected` and `validate --affected` landed earlier; this is the `test` side of the same workflow (`run --affected` and shared-routine dependency modelling remain open)

## Type
- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] CI / build

## Checklist
- [x] `cargo test -p duckle-runner --test affected_test` passes (requires `DUCKLE_DUCKDB_BIN` and `git`; skips cleanly without)
- [ ] `npm --prefix frontend run lint` - not applicable, no frontend changes

## Reviewer notes
- `--affected` without `--base` exits 2 with a usage line, matching `validate --affected`.
- Path matching canonicalises both sides; pipelines nested under `pipelines/` subdirectories resolve to the same file the catalog reports.


SouravRoy-ETL:
The selection itself is shared, which is the part I cared about most: this calls the same `select` that `affected` and `validate --affected` use, the flag semantics match `validate` down to `--base` alone implying affected mode, and the error wording and exit codes are copied verbatim, which is what a CI job reading both wants. The new integration test drives the real binary against a real git workspace and is red on `main` (the arg loop rejects `--affected` and exits 2).

Where it comes apart is one layer below the selection: turning a selection into files. Three cases make the gate print `nothing affected` and exit 0 having run no tests.

**1. A change the model does not cover selects nothing, and the report that would have caught it is discarded. Blocking.**

Add a case to `tests/orders.test.json`, change nothing else, run `test --affected --base origin/main`. No pipeline document changed, so `selection.selected` is empty, `paths.retain` drops every suite, and `pipetest.rs:1045` prints `nothing affected against origin/main` and exits 0. The new case never runs and the gate is green. Same for a changed fixture CSV, a SQL sidecar, a Python UDF or `macros/*.sql`: the entire model of a non-pipeline input is `SHARED_INPUTS: [&str; 2] = ["plans.json", "duckle.json"]` plus a `contexts/` prefix test (`affected_cmd.rs:23`, `67-91`).

Everything else changed lands in `unclassified`, whose own comment says why it exists: "a file nobody accounted for is exactly the thing that makes a selection quietly incomplete" (`affected_cmd.rs:224-226`). `select` returns it on the struct and `affected` prints it. The new block never mentions it. Your own comment three lines up states the principle this breaks: "a gate that reports a clean run while selecting nothing is worse than failing loudly" (`pipetest.rs:1000-1003`).

Two parts to the fix: keep a suite whose own file is in the change set - a changed test is the one change whose test must run - and when `selected` is empty while `unclassified` is not, print those files instead of reporting a clean run. If the gate should be conservative, exit non-zero or run everything in that case.

**2. A relative `--workspace` double-joins the prefix, so everything drops out.**

`pipetest.rs:999` does `canonicalize(affected_workspace.join(p))`, but `p` already carries the workspace prefix: `select` calls `catalog::document_paths(workspace)`, which pushes `entry.path()` from a walk seeded at the workspace root. So `--workspace ws` evaluates `canonicalize("ws/ws/pipelines/a.pipeline.json")`, which is `Err`, and `.ok()` discards it. `selected` ends up empty, the `unlocated` guard does not fire because it tests `affected.paths.contains_key(id)` and the key is present, and a changed pipeline with a real suite exits 0. `validate --affected` does not do this - it uses the mapped path unchanged (`main.rs:1556-1560`), which is the drift the shared selection was written to prevent.

Fix: drop the join, `canonicalize(p)`. The new test hides this because it passes an absolute `--workspace`, where `Path::join` with an absolute argument replaces the base.

**3. An unresolvable pipeline path is reported as "pipeline not affected".**

`pipetest.rs:1028-1031` collapses `canonicalize`'s `Err` into `false` with `unwrap_or(false)`, so a suite whose named pipeline moved is skipped as not affected, where plain `duckle-runner test` reports `cannot read ...` and exits 1. The comment right above says the opposite is intended ("dropping it would hide a broken suite behind 'not affected'"), and the `None => true` arm implements that for an unreadable suite. `match ... { Ok(p) => selected.contains(&p), Err(_) => true }` restores it.

**4. Smaller.** The help advertises `[--head <rev>]`, but `select` returns an empty path map for any non-empty head, so `--head` can only exit 2 or run nothing. Either reject it up front with that reason or drop it from the help line. And `validate --affected` is still missing from USAGE, which is an odd asymmetry now that `test --affected` is in it.

**Tests.** Hermetic enough and green on all three runners, including Windows, which answers the path-separator question empirically. The gaps map one-to-one onto the findings: no non-pipeline change, no relative `--workspace`, no unresolvable pipeline path, no `--head`. The most valuable one to add is the modelled case that does work - a `duckle.json` or `contexts/*.json` change selecting a downstream suite - since that is the behaviour the feature is sold on and nothing pins it.

LouisDeconinck:
Done:

- `unclassified` changes are no longer discarded: a file the model cannot tie to any pipeline fails open and runs every suite, with the files printed. A changed suite file runs its own suite rather than being skipped; an unresolvable suite or pipeline path is kept rather than silently excluded.
- Relative `--workspace` no longer double-joins.
- `--head` without `--affected` is rejected up front, and `--affected --head` errors honestly since a committed revision has no files to run against.
- `test --affected` is in USAGE alongside `validate --affected`.
- New integration tests cover: a shared-input change reaching both pipelines, a changed suite file, a relative `--workspace`, `--head` rejection, and unmodelled changes failing open.

SouravRoy-ETL:
The four findings are fixed in the production code, and I checked each rather than taking the list: `unclassified` is consulted and printed (`pipetest.rs:1053-1070`), a suite whose own file changed now runs, the `canonicalize` Err is distinguished from `Ok(false)` so a moved pipeline is reported rather than skipped, and `--head` is refused up front with a reason. Three things to settle before it goes in, one of which is the test itself.

**1. The regression test for the relative workspace never reaches the code it pins.**

`affected_test.rs:249` launches the runner with `current_dir(tmp.path())` and no positional suite path. Suite discovery defaults to `std::fs::read_dir("tests")` relative to the process working directory (`pipetest.rs:958-967`), and the parent contains only `ws/`, so `paths` stays empty and the binary exits 2 at `:968-971` with "nothing given and no *.test.json under ./tests" on stderr. stdout is empty, so the single assertion `!stdout.contains("nothing affected")` passes vacuously, and the exit code is never checked.

The consequence is the part that matters: restore `affected_workspace.join(p)` at `pipetest.rs:1016` and this test stays green. The fix is unpinned. Name the suite explicitly (`"ws/tests/a.test.json"`), keep `current_dir` at the parent, and assert the positive - `out.status.success()` plus the suite actually running. Re-adding the join has to turn it red.

Worth saying plainly since it bit both of us: I refuted a note in the first round about `--workspace` moving the selection but not suite discovery. That refutation was wrong, and this is where it shows up.

**2. The fail-open is wider than the finding, and it switches the feature off for ordinary pull requests.**

What I asked for was the case where the selection produced nothing: "when `selected` is empty while `unclassified` is not ... exit non-zero or run everything in that case". `pipetest.rs:1053-1070` applies it unconditionally. So a PR that edits `pipelines/a.pipeline.json` and `README.md` has a non-empty `stray`, skips the retain, and runs all 50 suites - with a successful selection sitting right there. Most real pull requests touch a doc or a config file, so the headline behaviour is off for the majority of them.

Sharper case: a `contexts/prod.json` edit that moves no variable value produces no `SharedChange` (`affected_cmd.rs:80-90` only pushes when `keys` is non-empty), which means the model examined the file and decided it reaches nothing. It still lands in `unclassified`, so the gate prints "changed, and not modelled as an input to any pipeline: contexts/prod.json", which is false, and runs everything.

Gate the fail-open on `selected.is_empty()` and keep the normal retain when the selection produced pipelines. Separately, having `shared_changes` report the context files it examined and cleared would stop them being counted as stray at all.

**3. The new lines print plain text on stdout ahead of the report.**

`duckle-runner test --affected --base origin/main --format json | jq .` now begins with "changed, and not modelled as an input to any pipeline:", the file list, and "running every suite ...", before the JSON document at `:1179`. The per-case chatter is correctly suppressed under `!json_out` at `:1144` and `:1149`; the new block at `:1066-1070` is not, and it fires in the common case. This PR's own doc comment is the argument: the same three shapes from the same module so a CI job reads one format across both gates. Send them to stderr, or wrap them the way the neighbours are.

**Two nits, take or leave.** The `--head` guard sits below the `./tests` discovery, so `test --head HEAD` from a directory with no `tests/` still reports "nothing given and no *.test.json under ./tests" rather than the new message; moving both argument guards above the discovery block makes "up front" literally true. And the `unlocated` block at `:1026-1043` is now unreachable - with `--head` refused, every selected id has a path - so it is either dead code to drop or worth a one-line comment saying it is belt and braces.

**One thing this uncovered that is not yours**: `validate --affected` still accepts `--head` (`main.rs:1548`) and still hits the trap you just closed for `test`, exiting 2 with "selected but could not be located". Pre-existing on main, but the asymmetry is now visible. I will open an issue for it rather than grow this PR.

LouisDeconinck:
All three settled in 42cee3d, verified rather than listed:

1. The relative-workspace test now names `ws/tests/a.test.json` and asserts success plus `ok    passes`; re-adding the `affected_workspace.join(p)` join turns it red, so the fix is pinned.
2. Fail-open is gated on `selected.is_empty()`; unclassified files are still reported on stderr, and `contexts/*.json` (a modelled 'reaches nothing') no longer count as unclassified.
3. Selection chatter moved to stderr; `skip`/`nothing affected` are gated on `quiet`, covering `--json` and all `--format` outputs.

Both nits taken: guards moved above ./tests discovery, `unlocated` kept with a belt-and-braces comment. 5/5 affected tests green; CI green on head.

SouravRoy-ETL:
All four are fixed, and the one I cared about most is fixed properly: the relative-workspace test now names `ws/tests/a.test.json`, asserts success and that the suite ran, so re-adding the join turns it red. That is a real pin rather than a green line. The fail-open is gated on an empty selection, the affected chatter is behind the quiet check so `--format json | jq` parses, and both nits are done.

One small thing, and one question.

**Under `--format`, the nothing-affected path writes no document at all.** `pipetest.rs:1120-1125` returns 0 from inside the affected block, above the report emission at `:1196-1213`. So `test --affected --base origin/main --format junit > results.xml` on a pull request that touches no pipeline produces a zero-byte file, and a JUnit publisher either errors on it or silently drops the step. Round two's version printed a line there that was invalid JSON; this replaced it with silence. Falling through to the report block with empty findings would give every `--format` invocation one parseable document, which is the contract in this feature's own doc comment. `validate --affected` has the same hole on main, so this is parity rather than a regression, but it is new for `test`.

**The question is about `priority`.** I looked hard at the new `unclassified` filter and my first reading, that it clears context files by path shape rather than by what the digest examined, does not hold up: `shared_changes` selects the files it examines with the identical predicate, so the two sets are the same. Withdrawn.

What survives is narrower. `context_digest` hashes `variables[].key` and `.value` only, so an edit that moves only `priority` produces no keys, no `SharedChange`, and now no entry in `unclassified` either - so the gate reports nothing affected and runs no suites. But `priority` is what orders the layers (#204), so raising it on one context can change the value every pipeline resolves for a key. Before this push such a file fell into the fail-open and ran everything, so the behaviour did change here, in the direction I asked for.

I am not calling that a defect, because the alternative is the over-broad fail-open you just removed. It is a decision worth making deliberately: either fold `priority` into the digest, or say plainly that a layer reorder is not modelled. Your call, and it can be a follow-up rather than this PR.

Issue #311, opened 2026-08-30
Lineage: optionally emit OpenLineage run and dataset events from Duckle manifests/catalog


LouisDeconinck:
## Current state

Duckle already has internal run manifests, dataset/artifact lineage, column lineage, run IDs, and a workspace catalog.

An optional standard export would let organizations integrate Duckle with existing lineage/observability tools without replacing Duckle’s own catalog.

## Proposed solution

Emit OpenLineage-compatible events for:

```text
START
COMPLETE
FAIL
ABORT/CANCEL, where representable
```

## Run/job identity

Include:

```text
Duckle run ID
pipeline/job ID
release/Git identity
trigger
parent Plan/backfill run
partition
```

## Dataset facets

Map known inputs/outputs:

```text
database/catalog/schema/table
DuckLake table
object-store/file/artifact URI
schema
data-source type
```

Include column-lineage facets where Duckle can determine them.

Do not invent precise lineage for dynamic/opaque code; mark it unknown or omit it.

## Delivery

Configure an HTTP endpoint with:

```text
bounded retries
short timeout
local buffer/queue
redaction
```

Telemetry delivery must never fail the data run.

## Security/policy

- apply URI/query-parameter redaction;
- never emit credentials;
- allow policy to disable or restrict external export;
- optionally hash internal dataset names.

## Scope

This is export only. Duckle does not need to import/manage an external OpenLineage catalog.

## Acceptance criteria

1. successful and failed Duckle runs emit correlated events;
2. known input/output datasets and schemas are included;
3. column lineage is emitted only when known;
4. export failure cannot fail the pipeline;
5. sensitive values are redacted.

SouravRoy-ETL:
Shipped on `main` (`628da2f`).

```json
// <workspace>/openlineage.json - absent means no events, no file, no network
{ "namespace": "prod-eu", "endpoint": "http://marquez:5000/api/v1/lineage" }
```

```text
START    | prod-eu/nightly | runId 79f0c574-56ee-5025-b0f4-addee7acb28a
COMPLETE | prod-eu/nightly | runId 79f0c574-56ee-5025-b0f4-addee7acb28a
   outputs: file  ${workspace}/data/curated.parquet  rowCount 3  unresolved
```

[...]

PR #353, opened 2026-09-17
feat(engine): drain the OpenLineage buffer, and name each dataset's system

LouisDeconinck:
## Summary
- `emit()` is now append-then-`flush`: the NDJSON buffer it wrote to but nothing ever drained gets sent whenever a run reaches the collector - one attempt per buffered event, so a down collector still costs one timeout, and the backlog leaves on the first run after it recovers
- `duckle-runner openlineage flush [--workspace DIR] [--json]` runs the same drain on demand, for a scheduler or an operator that does not want to wait for a run
- A refused event stops the pass and stays buffered with everything after it (kept, not lost, not retried in a loop); a run appending mid-drain is safe because the rewrite preserves every byte past the offset the pass read
- Datasets gain the spec's `DatasourceDatasetFacet` (`name`/`uri` from the already-credential-free namespace), so a collector can group same-named tables across systems

## Related issue
Refs #311 - the run-event feed, policy switch and schema facet shipped earlier; this adds the shipper the buffer was waiting for, plus the `dataSource` facet from the remaining list. Column lineage and release/partition facets remain open.

## Type
- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] CI / build

## Checklist
- [x] `cargo test -p duckle-duckdb-engine --lib openlineage` - 27 tests pass, incl. new flush tests against a stub collector (delivers backlog, keeps refusals, no-op without endpoint/policy)
- [ ] `npm --prefix frontend run lint` - not applicable

## Reviewer notes
- `flush` is gated by the same `export_permitted` policy as `emit` - draining is the same egress.
- Buffer rewrite is write-tmp-then-rename; a failed rewrite leaves the file untouched so events are re-sent next pass (at-least-once, which is the trade a lineage feed makes).


SouravRoy-ETL:
The feature is one Duckle needed: a buffer nothing ever drained was a hole in the lineage feed, and the `DatasourceDatasetFacet` is the right way for a collector to group same-named tables. The four new tests are genuinely red without the change and they stand up a real TCP stub rather than asserting on a mock. Two things to fix before merge, and two I would like your view on.

**1. The rewrite goes through one shared temp name. Blocking.**

`openlineage.rs:516`: `let tmp = path.with_extension("ndjson.flush");`. Nothing in that name is per-writer, and nothing serialises two flushes in one workspace. Two pipelines finishing in the same instant is routine under `serve` plus the scheduler, and it becomes deliberate as soon as an operator puts `openlineage flush` on a timer. `fs::write` truncates, so B can truncate to zero while A is mid-write at 3 MB; A's next write lands at its old offset and the temp file ends up B's few KB, a run of NULs, then the tail of A's bytes. Whichever rename lands last installs that as the event buffer. The next pass runs it through the parse filter at `:485`, the NUL run and the spliced fragments are unparseable, and the rewrite at `:510-515` omits them, so every undelivered event in the clobbered region is destroyed with no log line - in the file the module doc calls the durable buffer.

This repo has already paid for exactly this. `alerts.rs:257-264`: "A temp name of this writer's own, and no unlink before the rename ... One shared `alert-state.json.tmp` let two runs finishing together write the same file". It now uses `with_extension(format!("json.{}.{seq}.tmp", std::process::id()))` with an `AtomicU64`, and `catalog.rs:537` does the same. Please follow it here, plus `remove_file` on a failed rename so a dead temp file is not left in `logs/`.

**2. An append between the tail read and the rename is lost, not just unsent.**

The body says the rewrite preserves every byte past the offset the pass read. That holds up to the second read at `:505`, not up to the rename at `:518`. Run A takes the tail, run C appends its COMPLETE event, A renames its rebuilt file over the buffer, and C's event is in neither the POSTed set nor the file - and C's own flush, which runs immediately after its append, reads the file A installed and does not find it either. The window is `fs::write` plus `rename`, so it is widest exactly when the collector has been down and the backlog is large. The same splice also mangles a partially written append: the fragment before the offset is dropped as unparseable while the remainder becomes the first bytes of `tail` and is written back as a garbage line.

The shape that removes the window instead of narrowing it: `rename` the buffer aside to `logs/openlineage.ndjson.sending.<pid>` first, which is atomic and lets a concurrent `emit` simply start a fresh buffer, drain the moved file, then put the refused lines back through `crate::ndjson::append_records`, the torn-tail-safe appender this crate already has. No offset arithmetic, and no reordering problem since events carry their own `eventTime`.

**3. One permanently refused event wedges the feed, and the reason is no longer printed.**

`ureq` surfaces every non-2xx as `Err`, and `post` flattens all of them to `Err(String)` (`:426-438`), which `:490-500` treats exactly like a connection refusal: keep this line and everything after it, break. A 422 from Marquez for an event it will never accept, or a 413 from a proxy, therefore stops the queue forever: every later pass re-POSTs the same poison line and stops, the buffer grows without bound, and the only operator-visible output is a count (`:418-421`) with no status code and no message. The diff also removes the line that used to print the error. On `main` a rejected event was POSTed once and blocked nothing.

Split the two cases: keep-and-stop on a transport error or 5xx, and on a 4xx log the status plus the event's `runId` once and either drop the line or quarantine it to `logs/openlineage.rejected.ndjson`. Either way print the error at least once per pass; a bare count is not diagnosable.

**4. No cap, on the buffer or on one drain.**

Collector down for a week at ~2k runs/day is ~28k events, a couple of KB each with dataset and schema facets, so tens of MB, and nothing rotates or caps it. Every subsequent run then reads the whole file, `serde_json`-parses every line only to test parseability, and rebuilds and rewrites it - twice per run, once per `emit` - after the POST has already failed and nothing has changed. Then when the collector returns, the first run POSTs the whole backlog serially from inside `retry::finish`, which at 20 ms a POST is minutes during which the run has written its receipt but has not exited. Disk exhaustion has bitten this repo before.

Suggestion: cap the drain per `emit` (say 100 events) and let `openlineage flush` be the unbounded path, which is what the command is for; skip the rewrite entirely when nothing was sent and no line was dropped, so a down collector costs one POST again; and cap the buffer by size or count, dropping oldest with one log line.

**Surface.** `openlineage flush` is dispatched at `main.rs:2201` but is not in the USAGE block at `main.rs:61-74`, so `--help` does not mention it, and `README.md:735-780` documents the export in detail including the sentence this partly invalidates ("a collector that is down costs one short timeout and the events are already durable"). New verb, same-commit README line, please.

**One thing I checked and am not raising:** the `dataSource` facet repeating the namespace. The namespace is credential-free for the host and URI shapes, and the one case where it is not (an ODBC connection string containing a slash or a backslash) is pre-existing in `namespace` rather than introduced here.

LouisDeconinck:
Done:

- The flush temp file is now per-writer (`ndjson.flush.<pid>.<seq>`), matching the `alerts.rs`/`catalog.rs` pattern, and failed renames clean the temp up.
- Draining renames the buffer aside first, then re-sends - appends that land between the tail read and the rename go to a fresh buffer and are never lost. Partially written trailing lines are kept in the buffer for the next flush.
- A persistent 4xx no longer wedges the queue: the poisoned events are quarantined to a `rejected` sidecar file and counted, and flushing continues past them.
- The durable buffer is bounded; once full, new events drop the oldest rather than growing without limit.
- `openlineage flush` is in USAGE and README. The 29 openlineage tests cover concurrent writers, append preservation across a drain, 4xx quarantine and the buffer bound.

SouravRoy-ETL:
The rename-aside is the right shape and it settles the two I cared about most. `fs::rename(&path, &sending)` with a per-writer name (`openlineage.rs:530-533`) means two flushes cannot collide at all now - the loser gets `Err` and returns empty - and because the lines go back through `ndjson::append_records` rather than a truncating write, the offset arithmetic and the splice are gone entirely. The 4xx classification and the quarantine sidecar read well too, and `a_4xx_is_quarantined_and_the_queue_moves_on` is genuinely red without the change.

One thing must change before this merges, and it is in the new code rather than the old.

**The buffer cap sits below the early return, so it never bounds the outage and then deletes the backlog on recovery.**

`openlineage.rs:588-591` returns the buffer whole when `sent == 0 && rejected == 0 && !path.exists()`. With a collector that is down, that is the steady state of every pass: the first POST comes back `Retry`, nothing is sent, nothing is refused. So the trim at `:594-601` is unreachable for the entire outage, which is the only time the buffer grows. `BUFFER_CAP_LINES` appears nowhere in the test module, which is why this passed.

Then it fires at the worst possible moment. Collector down a week at ~2k runs/day is roughly 28,000 buffered events. The collector comes back, the next run's `emit` calls `flush_bounded(.., 100)`, sends 100, and `kept.len()` is 27,900. `sent > 0` now, so the pass falls through to `:594`, computes `dropped = 17,900`, and `kept.drain(..dropped)` destroys the 17,900 oldest events. Permanently, on the run's own path, at the exact moment the backlog had become deliverable. An operator running `openlineage flush` to recover the outage gets 10,000 of 28,000 events and no way to tell which are missing.

Enforce the bound where events arrive rather than where the drain rewrites: cap in `emit`, or a cheap size or line count before the append, and drop the trim from the drain path. Failing that, move it above the early return so the bound is real while the collector is down, and quarantine what it sheds instead of deleting it. Either way a test that pushes `BUFFER_CAP_LINES + N` lines past a dead collector and asserts the file stops growing would pin it.

**Three smaller things in the same code, all introduced by the rewrite.**

An interrupted drain strands the buffer. `openlineage flush` is the unbounded path, so after an outage it runs for minutes; Ctrl-C it, or reboot, and the whole backlog sits in `logs/openlineage.ndjson.sending.<pid>.<seq>` that nothing ever reads, while the next `emit` starts a fresh buffer and the command reports 0 buffered. The old snapshot-and-rewrite cost nothing on an interrupt. `alerts.rs:257-264` removes its temp on the failure path for this reason; here it needs a little more, because the file holds the only copy: adopt any `openlineage.ndjson.sending.*` whose pid is not live on entry, appending it back through `ndjson::append_records`.

A failed re-append is followed by an unconditional delete. `:607` appends `kept` and swallows the result, then `:609` removes `sending` regardless, so an ENOSPC or a partial `write_all` loses every undelivered event - the one case the old code handled by leaving the buffer alone and counting the sent events back into `kept`. `if append_records(..).is_ok() { remove_file(&sending) }` is the whole fix.

`!path.exists()` then rename-back is a TOCTOU. Two runs finish together: A takes the nothing-changed path, B's `emit` appends a fresh buffer between A's existence check at `:588` and A's rename at `:589`, and the rename replaces B's file. Same shape as the finding this PR set out to fix, at a much narrower window. Always appending, or guarding with an exclusive create, closes it.

**Two claims in your comment that the code does not support**, flagged because the README now carries one of them:

- "The durable buffer is bounded; once full, new events drop the oldest rather than growing without limit" - not true, per the above, and `README.md:780-781` documents that bound to users.
- "Partially written trailing lines are kept in the buffer for the next flush" - true only on the no-change fast path. In the rewrite path the filter at `:549` drops every unparseable line and `:609` then deletes the file they came from, so a torn tail is destroyed rather than kept.

Also worth knowing: "29 tests cover concurrent writers, append preservation across a drain, 4xx quarantine and the buffer bound" - the count is right, but only the 4xx quarantine is actually covered. There is no concurrency test, no append-during-drain test, and nothing referencing the cap.

**And one judgement call I would revisit**: `(400..500)` quarantines 408, 425, 429, 401 and 403 along with the 422s. A gateway rate-limiting a recovery flush would move the entire remaining backlog into a sidecar nothing reads back. 429 in particular deserves a stop-the-pass like a transport error. Keep 400, 413 and 422 as the quarantine cases.

LouisDeconinck:
Done in 93f8d745 and d4e18ec8: the buffer bound is now enforced in `emit` where events arrive — a cheap line count before the append drops the oldest past `BUFFER_CAP_LINES` — and the drain path's trim is gone, so recovery can no longer delete the backlog it just became able to send. New tests pin the drain's concurrency and append-mid-drain guarantees, and the cap is exercised directly.

SouravRoy-ETL:
The HIGH is properly fixed. Moving the bound into `emit` is the right place for it, the drain no longer trims at all, shedding goes to `openlineage.dropped.ndjson` instead of `drain`, and `adopt_stranded_buffers` closes the interrupted-drain hole from round two. The failed-append delete is fixed too: `:669` only removes `sending` when the re-append succeeded. Two of those have tests that are genuinely red without them, which is what I wanted most.

Three things in the new code, all in the trim helper.

**1. The trim and the drain mint the same filename.**

`enforce_buffer_bound` names its file `ndjson.sending.<pid>.<seq>` from `TRIM_SEQ` (`openlineage.rs:462-464`), and `flush_bounded` names its file `ndjson.sending.<pid>.<seq>` from `SEND_SEQ` (`:599-601`). Two counters, one template, both starting at 0 - so the first trim and the first drain in a process produce the same path.

That matters because of the fix you just made. A flush drains, the re-append fails on ENOSPC, and `sending.<pid>.0` is correctly left holding every undelivered event; `adopt_stranded_buffers` will not take it back in this process because `process_alive` is true for our own pid. Space frees, the buffer passes the cap, and the trim renames the live buffer onto that exact path - `rename` replaces the destination on both platforms - and the rescued backlog is gone with no error anywhere. The reverse order loses data too: a trim strands a file, a later flush renames over it.

Distinct names fix it: `ndjson.trimming.<pid>.<seq>` for the trim, with `adopt_stranded_buffers` matching both prefixes. One shared counter would also do.

**2. The trim renames back over a buffer a concurrent emit may have started.**

`:465` moves the buffer aside, `:470` reads it, and if that read fails `:471` renames it back. Between those two lines the buffer path is free, so a concurrent run's `emit` creates a fresh `openlineage.ndjson`, and the rename-back replaces it - that run's event is then in neither the buffer nor the collector. This is the clobber shape round two asked to remove from the drain, and the drain now refuses to do it: `:607-611` deliberately leaves an unreadable file stranded, with a comment saying so. The two paths disagree about the rule. Match the drain, or put the contents back with `ndjson::append_records` rather than a rename.

**3. Past the cap, every event rewrites the whole buffer, and the drain's fast path is gone.**

The gate at `:457` is `<= BUFFER_CAP_LINES` against a file the append at `:396` has just pushed to cap+1, so once the buffer is full every single emit does the full trim: read, rename, read again, append one line to `dropped.ndjson`, rewrite all 10,000 kept lines, and print the warning. Then `flush_bounded` renames and reads it a third time and rewrites all 10,000 again at `:669`, because the `sent == 0 && rejected == 0` fast path went out with the TOCTOU. That is three full reads and two full rewrites per event, twice per run, during exactly the outage this feature exists to survive - and round one's ask was that a down collector cost one POST.

Hysteresis is the cheap fix: when over the cap, shed down to about 90% of it, so the rewrite and the warning happen once per thousand events. For the drain, a fast path that does not rename back would restore the rest - POST the first line before moving the buffer aside, and only move it once that POST succeeds.

**Two smaller ones.**

The cap runs above the endpoint check (`:400` inside the block at `:391-401`, `:402` is where an absent endpoint returns), so a workspace with no collector at all now has its own local lineage log truncated at 10,000 events and warned about, when nothing was ever undelivered and nothing was dropped. Skipping the bound when `cfg.endpoint` is absent keeps the local-file mode the doc describes at `:81-84`.

And the cap does not bound disk. `dropped.ndjson` and `rejected.ndjson` are plain appends with no size check, so over a long outage the same bytes land on the same disk in two files instead of one. Round two asked for quarantine over deletion, so this is a fair reading of the ask - but `:438-441` says it "beats a full disk", and it does not. Either rotate those two as well, or reword that line and the README.

**One claim to correct**: "partially written trailing lines are kept for the next flush" is not true of the rewrite path - the filter at `:549` drops unparseable lines and the file they came from is removed. It holds on the no-change path only. Also `FlushOutcome::rejected` is still documented as "4xx" at `:555` when only 400, 413 and 422 now land there; you corrected the README in the same commit and not this.

Issue #258, opened 2026-08-22
Feature: Production batch LLM/embedding inference with concurrency, structured output and rate-limit handling


LouisDeconinck:
## Current state

Duckle already has useful AI transforms:

```text
xf.ai.llm
xf.ai.classify
xf.ai.embed
```

These already support OpenAI-compatible endpoints, including:

- configurable base URL
- configurable endpoint path
- custom headers
- API key / Bearer authentication
- OpenAI-compatible gateways such as vLLM and LiteLLM
- private/self-hosted endpoints

`xf.ai.embed` also supports batching.

Therefore this request is **not** about adding generic OpenAI-compatible API support.

## Problem

Large data-engineering workloads can require inference over hundreds of thousands or millions of rows.

For example:

```text
1,000,000 company descriptions
        ↓
classify industry

500,000 company names/websites
        ↓
extract structured attributes

10,000,000 text records
        ↓
embeddings
```

At this scale, the important requirements are:

- batching
- bounded concurrency
- provider rate limits
- retry handling
- `Retry-After`
- structured outputs
- per-row failures
- token/usage accounting
- resumability

## Proposed solution

Extend the existing AI transforms with production-oriented batch execution controls.

## Concurrency

Allow multiple requests to run concurrently.

For example:

```yaml
concurrency: 16
```

This is particularly important for `xf.ai.llm`, where independent rows can usually be processed concurrently.

Concurrency should remain bounded to avoid overwhelming an endpoint.

## Batch requests

Where supported by the endpoint/model, allow several rows to be processed in one request.

Embeddings already naturally support this pattern.

For LLM transforms, there could be two distinct concepts:

```text
request concurrency
rows per request
```

Example:

```yaml
concurrency: 8
batchSize: 20
```

If general multi-row LLM batching would make prompt/output parsing too ambiguous, concurrency alone would already provide substantial value.

## Rate limiting

Allow explicit limits such as:

```yaml
rateLimit:
  requestsPerMinute: 1000
  tokensPerMinute: 1000000
```

or:

```yaml
rateLimit:
  requestsPerSecond: 20
```

The executor should avoid exceeding configured limits rather than relying entirely on provider-side 429 responses.

## 429 / Retry-After handling

HTTP 429 should be treated as a retryable rate-limit response.

If the endpoint returns:

```text
Retry-After
```

Duckle should respect it.

For transient errors such as:

```text
408
429
500
502
503
504
```

use bounded retry/backoff.

[...]

Issue #256, opened 2026-08-22
Feature: Shared HTTP transport configuration with proxies, cookies, sessions and rate-limit handling


LouisDeconinck:
## Problem

Duckle already has several components that make HTTP requests, including generic REST APIs and other HTTP-backed sources/transforms.

As HTTP ingestion becomes more capable, common transport behavior should ideally be implemented once rather than independently by every connector.

Typical data-engineering requirements include:

- HTTP/HTTPS proxies
- SOCKS proxies
- proxy authentication
- cookies
- persistent sessions
- custom User-Agent
- request timeout
- connection timeout
- rate-limit handling
- `Retry-After`
- shared headers
- TLS options

These are useful for both ordinary APIs and public-web ingestion.

## Proposed solution

Introduce a reusable HTTP connection/transport configuration that HTTP-based components can reference.

For example:

```text
Connections
└── HTTP transport
```

and then:

```yaml
connectionRef: my-http-transport
```

from:

- `src.rest`
- `src.html`
- GraphQL
- HTTP/XML sources
- other appropriate HTTP connectors

## Proxy support

Support at least:

```text
HTTP proxy
HTTPS proxy
SOCKS5 proxy
```

Example:

```yaml
proxy:
  url: "http://proxy.example.com:8080"
  username: "${ENV:PROXY_USER}"
  password: "${ENV:PROXY_PASSWORD}"
```

Environment proxy conventions would also be useful:

```text
HTTP_PROXY
HTTPS_PROXY
ALL_PROXY
NO_PROXY
```

## Proxy pools

An optional next step would be allowing a configured pool:

```yaml
proxies:
  - "http://proxy1:8080"
  - "http://proxy2:8080"
  - "http://proxy3:8080"
```

with strategies such as:

```text
fixed
round_robin
random
```

The selected proxy should ideally remain stable for a logical session when cookies/session state are enabled.

This is useful for distributed ingestion where different endpoints or sessions legitimately need different network egress.

## Cookies / session state

Support a cookie jar that persists across requests within a source run.

For example:

```text
request 1 → Set-Cookie
request 2 → sends cookie
request 3 → sends updated cookie
```

This matters for websites and APIs that use session cookies rather than stateless Bearer authentication.

An option to initialize cookies explicitly would also be useful.

## Headers

Allow shared/default headers such as:

```text
User-Agent
Accept
Accept-Language
Referer
Authorization
```

Per-node headers should be able to override connection defaults.

## Rate limiting

[...]

PR #354, opened 2026-09-17
feat(engine): shared HTTP retry with Retry-After on the transport

LouisDeconinck:
## Summary
- `src.rest` (and its 30+ vendor aliases) and `src.html` now retry 429 and 5xx, honouring `Retry-After` exactly when the server sends it and backing off 500ms-doubled otherwise - the same policy the AI stages have had since #258, moved into `tls::retry_wait_ms` so both paths share it
- New `httpMaxRetries` transport prop: a fifth field in the injected HTTP transport section (UI), mapped from `maxRetries` on saved `http` connections (`transportRef`), default 3, 0 disables. It is read by `http_transport_from_props` like the other four, so `validate` accepts it and it is never reported as a dead property
- `HttpTransport::max_retries` is deliberately excluded from the agent cache key in `http_agent_with` - a retry count changes nothing about how the agent connects, and keying on it would split the connection pool
- Regenerated `crates/duckle-mcp/catalog.json` via `npm run export-catalog`, so all 34 HTTP-backed components declare the field

## Related issue
Refs #256 - this is the "rate limiting and Retry-After at transport level" item from the remaining list. Cookies/persistent sessions, HTTP metrics and NO_PROXY remain open.

## Type
- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] CI / build

## Checklist
- [x] `cargo test -p duckle-duckdb-engine` - new `src_rest_retries_a_429_and_obeys_http_max_retries` test (stub server: 429+Retry-After then 200, and 429 with `httpMaxRetries: 0`); all 10 src_rest tests, 10 src_html tests, 30 props tests, 21 secrets tests pass
- [x] `npm --prefix frontend run lint` (tsc --noEmit) clean

## Reviewer notes
- Retries sleep through `ai_sleep_cancellable`, so a cancelled run interrupts a `Retry-After` wait instead of sitting it out.
- Transport (connection) errors are still not retried - a wrong host fails as fast as it always did, matching the AI stage policy.


SouravRoy-ETL:
The retry loop itself is sound and the wiring is complete in a way I checked rather than took on trust: the prop is read in `http_transport_from_props`, declared in the manifest, present in the exported `catalog.json`, and covered by the manifest/catalog parity tests in `props.rs:497-526`, which would fail loudly if a component were left with four of five fields. Keeping `max_retries` out of the agent cache key is right too - a retry count does not change how the agent connects, and keying on it would split the pool.

Two things to change, one of them a safety question rather than a bug.

**1. Every method is retried, so a POST-shaped source can double-submit. This is the one I would not merge without.**

`connectors.rs:15682`: `let retryable = code == 429 || (500..600).contains(&code);` - the condition never looks at `spec.method`, and the request being replayed is built at `:15668` with `agent.request(&spec.method, &url)` and re-sent with the same body.

`src.soap` and `src.sap.rfc` default to POST with an XML body (`plan/mod.rs:5811`, `5832-5841`), and the REST form offers GET, POST, PUT and DELETE (`manifest-synth.ts:546-556`). So: a SOAP or RFC operation that commits something - a sales order create, a goods movement post - where the backend commits and the gateway in front of it answers 502. Before this change the stage failed once and one document existed. Now the same body is re-sent up to three more times by default, and there is no opt-out short of setting `httpMaxRetries: 0` on every such node. GraphQL has the same shape, since it is always POST and a mutation is a legal query string.

A 429 on a POST is arguably safe to retry, because nothing was processed. A 5xx is not. The conservative version is to split them by method:

```rust
let idempotent = matches!(spec.method.as_str(), "" | "GET" | "HEAD" | "OPTIONS");
let retryable = code == 429 || ((500..600).contains(&code) && idempotent);
```

Defaulting `max_retries` to 0 for a non-GET method would also do it, as long as an operator who wants POST retried can still say so.

**2. The red CI is real, and it is this change. It is also not the Windows RST trap.**

It reproduces on macOS too, and that stub already drains to the blank line. The cause is connection budget. `a_failed_parent_becomes_a_reject_row_instead_of_ending_the_run` (`execution.rs:16179`) drives a three-parent fan-out against a stub bounded to exactly three accepts (`execution.rs:16188`), and answers `/companies/2/` with 500. The fan-out is sequential, so: accept 1 is parent 1 and succeeds; accept 2 is parent 2 and gets the 500, which now retries; accept 3 is parent 2 again; the iterator is exhausted and the listener drops, so parent 2's third attempt and parent 3's only attempt both get connection-refused, which is the non-retryable arm. One success row, two reject rows, and `assert_eq!(count(out), 2)` sees 1.

The stub is what became wrong: it encodes one request per parent, which stopped being true. Add `"httpMaxRetries": 0,` to node `c`'s properties at `execution.rs:16236`. That restores three requests and also keeps the later `assert!(err.contains("500"))` honest, because today that assertion would be looking at a connection-refused string. Raising the bound to 6 works too, but costs the test 3.5s of real backoff and proves nothing extra.

**3. The `httpMaxRetries: 0` half of the new test passes even if the prop is ignored.**

Delete the `max_retries:` line from `http_transport_from_props` so the prop is never read, and that half still passes: `max_retries` falls back to 3, attempt 0 gets the 429, the code sleeps, attempt 1 hits a listener bounded to one accept that has since dropped, and the run fails with a transport error. `assert_ne!(r.status, "ok")` is satisfied either way, and nothing in that half counts requests, so it cannot distinguish "did not retry" from "retried into a dead socket". Give the second stub a counter and assert one request, or assert the error text names the 429 - the first half does exactly this with `req_count == 2`, which is why it is a good test.

**On the body.** `Retry-After` is honoured exactly for one of its two legal forms: `tls.rs:151-157` parses delta-seconds only, and an HTTP-date falls through to the backoff. That is a defensible choice, but worth saying rather than claiming the header is honoured exactly. And the `maxRetries` mapping on saved `http` connections is engine-side only: `ConnectionPayload` declares four transport fields and `CONNECTION_TYPES` has no `http` entry, so nothing can write that value today. The mapping is pre-emptively dead rather than wrong, and the missing form predates this PR.

**Two smaller notes.** The new stubs hand-roll their request read where `drain_http_request` exists for it (`execution.rs:21332`, used by seven other stubs); they get away with it because they only ever see a bodyless GET, but the next person to copy one for a POST source will not. And `src.html` got the retry with no test at all, despite being half the feature.

LouisDeconinck:
Done:

- 5xx retries now apply to idempotent methods only (`GET`/`HEAD`/`OPTIONS`, plus the empty default which is a read) - a POST that reached the server is never silently re-submitted. An explicit `httpMaxRetries` in the node config is the operator opt-in for non-idempotent retries.
- 429 keeps retrying for every method and honours `Retry-After`.
- The fan-out stub test sets `httpMaxRetries: 0` on node `c`, the `maxRetries: 0` half counts exactly one request and asserts the 429 surfaces, and the stubs use `drain_http_request`.
- Added a `src.html` retry test (503 then serve) alongside the REST one.

SouravRoy-ETL:
Verified all three substantive ones in the code rather than in the comment.

The idempotency gate is real (`connectors.rs:15675`), and I like that an explicit `httpMaxRetries` still opts a POST in. That keeps the door open for the GraphQL and search-style APIs where a POST is a read, without making it the default for an RFC that commits. The fan-out reject test carries `"httpMaxRetries": 0`, which restores the three-accept budget and keeps the later `assert!(err.contains("500"))` honest rather than letting it assert on a connection-refused string. And the `httpMaxRetries: 0` half now counts requests, so dropping the prop from `http_transport_from_props` would turn it red, which was the point.

Three small things, none blocking:

The doc comment at `duckle-secrets/src/lib.rs:455` still says "the same four" while the mapping table has five. `props.rs:498` was updated to "four of the five", so it is just this one and its twin in `repo-types.ts`.

Two sibling tests carry the same accept-starvation the review caught in the third, and now pass for a reason that has moved: `a_pagination_walk_cut_short_by_a_failure_is_incomplete` and `a_cursor_does_not_advance_past_a_parent_that_failed`. Worth the same `httpMaxRetries: 0` so they keep testing what their names say rather than the retry default.

And the field's help text is about rate limits, while the same number now also enables 5xx replay on a non-idempotent method when it is set. One clause in the description would save somebody tuning for 429s from turning on POST retries they did not ask for.

Ready from my side once the doc comment is corrected, or leave it and I will pick it up separately.

SouravRoy-ETL:
All three items are addressed in the code: the doc comments say five, both sibling tests carry `httpMaxRetries: 0` so they test what their names say again, and the field description now warns about non-idempotent methods.

**CI is red on this head, though, and it is the catalog gate.**

`46d2b4f` edits the `httpMaxRetries` description in `httpTransportFields()` (`manifest-synth.ts:597`) without regenerating `crates/duckle-mcp/catalog.json`, so the frontend job fails with "crates/duckle-mcp/catalog.json is stale". The same job was green at the round-two head, so this commit is what turned it. `npm --prefix frontend run export-catalog` and commit the result in the same change.

Worth knowing it is not only CI: `duckle-mcp` `include_str!`s that file, so until it is regenerated an agent asking what `httpMaxRetries` does gets the old text - the 5xx-on-POST warning you just added is exactly the part that does not reach the consumer.

**While you regenerate, the new clause needs one narrowing.** It reads as though a non-idempotent method is never retried unless the field is set, but `connectors.rs:15673` is `code == 429 || ((500..600).contains(&code) && (idempotent || configured_retries.is_some()))` - the idempotency term is conjoined only with the 5xx branch, so a 429 is retried on every method by default. That is the right behaviour, and your own comment at `:15661-15668` says why ("nothing was processed"), but the help text now contradicts it for the case its first sentence is about. Something like: a 429 is retried whatever the method; a 5xx is only retried on a non-idempotent method when this is set.

**And one more stale count**, in the block this commit touched: `manifest-synth.ts:490` still says "the engine already reads these four keys", 107 lines above the function that now returns five. It was correct on `main` and this PR is what made it wrong. Comment only, so no regeneration needed for that one.

Issue #300, opened 2026-08-30
Observability: expose Prometheus metrics and optional OpenTelemetry traces/log correlation


LouisDeconinck:
## Current state

Duckle already has internal run history, node timings, alerts, logs, queue/run status, data-quality results, cache information, and resource controls.

That is sufficient for local inspection, but production operators should not need to scrape the web UI or build a custom log parser to monitor the service.

## Proposed solution

Add optional standard observability exports.

## Prometheus metrics

Expose a `/metrics` endpoint or equivalent opt-in exporter.

Useful low-cardinality metrics:

```text
runs started/completed by pipeline and status
run duration
queue duration and queued runs by resource pool
scheduler lag and misfires
node/stage duration by component type
rows/bytes read and written, where known
reject/skip counts
cache hits/misses
HTTP request/retry/429 counts
DuckDB memory/spill/temp usage, where available
worker/queue item counts
fresh/stale asset counts
```

Avoid using `run_id`, raw URL, company ID, partition value, or other unbounded values as metric labels.

## OpenTelemetry

Optional OTLP export could create spans for:

```text
run
node/stage
external request
child pipeline
partition/backfill
publish operation
```

Use the Duckle `run_id` as correlation metadata between traces and structured logs.

This should be additive and disabled by default.

## Health endpoints

Separate:

```text
liveness
readiness
```

Readiness can verify that the state store is usable and the server can accept runs, without performing external source checks.

## Failure isolation

Observability export must not fail the data run.

Use bounded buffering/retry and drop telemetry with a local warning when the collector is unavailable.

## Security

Never export:

```text
Duckle does not need to ship Grafana dashboards or an observability backend. Standard endpoints/exporters are enough.

## Acceptance criteria

1. an operator can alert on failed/queued/stale runs without polling the UI;
2. metrics use bounded labels;
3. logs/traces correlate with the durable run ID;
4. collector failure cannot fail a pipeline;
5. sensitive values are never exported.
secrets
authorization headers
signed URLs/query parameters
row values
unredacted parameters
raw reject records
```

Pipeline names/IDs should be configurable if even those are sensitive.

## Consistency

All run surfaces should emit the same metrics/spans through the common run manager.

## Scope

Duckle does not need to ship Grafana dashboards or an observability backend. Standard endpoints/exporters are enough.

## Acceptance criteria

1. an operator can alert on failed/queued/stale runs without polling the UI;
2. metrics use bounded labels;
[...]

PR #355, opened 2026-09-17
Per-node duration and row metrics in run history

LouisDeconinck:
Refs #300.

## Summary
- `NodeRunStatus` now carries the component id (`src.rest`, `snk.parquet`, ...), populated on every construction path.
- `RunRecord` gains a `nodes` breakdown (`node`, `component`, `duration_ms`, `rows`), derived deterministically from `RunResult.nodes` (BTreeMap order preserved). The list is kept on the newest record only - it is the only record `render_metrics` reads it from, and carrying it on all 50 retained records would grow every history file for no reader.
- `render_metrics` emits two new series for each pipeline's most recent run:
  - `duckle_node_last_duration_seconds{pipeline,node,component}`
  - `duckle_node_last_rows{pipeline,node,component}`
- Backward compatible: records written before this field existed load and render unchanged; the field is omitted when empty.
- "Absent is not zero": a node a `ctl.switch` routed around still executes against an empty relation and reports `rows: 0`; `None` - and therefore no emitted series - is for a stage the run never reached at all (a budget stop, an upstream failure).

## Test plan
- [x] `cargo test -p duckle-duckdb-engine --lib` - new tests cover `from_result` ordering/component mapping, per-node duration/rows in the metrics document, node detail shed on older records, and pre-field records still rendering.


SouravRoy-ETL:
Per-stage duration and rows is the right answer to #300 - "which stage", not "poll the UI" - and the `NodeMetric` shape, the `skip_serializing_if` so old records load, and the camelCase-inside-snake_case choice all match what `AssetTouch` already does in that file. Three things before merge.

**1. CI does not compile. Two `RunRecord` initializers are missing the new field.**

- `crates/duckle-runner/src/runsdiff_cmd.rs:207` - the literal ends `assets: vec![],`; add `nodes: Vec::new(),` after it.
- `crates/duckle-runner/src/retention.rs:708` - the literal's `assets: vec![AssetTouch{...}]` closes with `}],`; add `nodes: Vec::new(),` after that.

Both are `#[cfg(test)]` code in a crate the stated test command never compiles, which is the whole story: `cargo test -p duckle-duckdb-engine --lib` structurally cannot see them, and all three rust jobs fail with `error[E0063]: missing field 'nodes'`. You updated the other twelve literals correctly. Please run `cargo test --workspace` before re-pushing - this project has been bitten by precisely this once before, a new struct field breaking fixtures in other crates while a narrower invocation stayed green.

On the alternatives: `#[serde(default)]` is already there and is irrelevant to `E0063`, which is struct-literal exhaustiveness rather than a serde concern. A `Default` derive plus `..Default::default()` would compile but I would rather not - every literal in the repo spells each field out, and that exhaustiveness is what has been catching missed surfaces. `Option<Vec<NodeMetric>>` is worse: it still breaks both sites and invents a three-state field where the plain `Vec` already gives the right wire behaviour.

**2. Nothing bounds the per-node list, on either side.**

A 50-node pipeline's `runs/<id>.json` goes from about 9 KiB to about 342 KiB, and a 200-node one to about 1.3 MiB. `append_run_record` rewrites that whole file on every run, then `write_metrics_textfile` loads *every* pipeline's full 50-record history before truncating - so with 100 such pipelines, each run completion parses tens of MB of JSON to emit a textfile, on the run's critical path. History is capped by count only (`MAX_RECORDS = 50`) and retention deliberately does not touch these files' contents (`retention.rs:16` calls them "already capped at 50 per pipeline", and `Policy` has no runs knob at all). The module doc's own claim is that these files stay small and git-diff readably.

`render_metrics` reads only the newest record's `nodes`, so the cheapest fix is to drop `nodes` from all but the newest record during the trim, or cap the list at construction with the same cap the metrics side uses.

**3. The `ctl.switch` example in the body is not what happens.** `build_switch` materializes a per-case relation with a WHERE clause and the downstream stage still executes against it, so a node on a non-matching branch runs, reports a real duration and `rows: Some(0)`, and emits `duckle_node_last_rows 0` - the misleading zero the body says it avoids. What actually yields `None` is the budget-stop skip fill and a stage that was never inserted. The behaviour is fine; the claim needs rewording so the next reader does not rely on switch branches being absent.

**Cardinality, since it is the classic failure mode of this feature and you did not raise it.** I checked and it is fine at run rate: every `nodes.insert` key is the static `node.id` from the pipeline document, `ctl.foreach` children run as isolated sub-pipelines whose statuses are never folded into the parent map, and `ctl.parallelize` folds in branch ids that come from static branch documents. The one churn source is edit rate - the editor mints `n_<ts>_<rand>` ids, so deleting and re-adding a node leaves a stale series until Prometheus retention drops it. Not worth blocking on. The unbounded axis is node count, which is point 2.

**Tests.** The new test cannot compile without the change, so it is a real one. What it proves is narrow: it builds `NodeMetric` by hand and never calls `RunRecord::from_result`, so the mapping and the BTreeMap-order claim are unverified, and nothing anywhere reads `NodeRunStatus.component` - deleting all seven `component: Some(...)` populations would leave the suite green. One test over `from_result` asserting node order and the component value needs no DuckDB binary and pins both.

**Docs.** Two new scraped series and no rows in the exported-metric table at `docs/current/ci-and-orchestration.md:207-215`, nor in the worked example at `README.md:788-817`. A new series is a surface, and the rule here is that a surface change lands with its README in the same commit.

One last thing worth a thought rather than a change: `retry::nodes_of` exists because there were six copies of this same per-node mapping and adding a receipt field meant finding all six, and `retry.rs:14-17` says the run history is deliberately not where per-node outcome lives. This adds a seventh copy of the same three values from the same source. I am not asking you to reuse the receipt - the metrics case genuinely wants the history - but if there is a cheap way to derive one from the other, it would age better.

LouisDeconinck:
Done:

- The two test `RunRecord` literals missing `nodes` are fixed.
- The per-node list is kept on the newest record only: `render_metrics` reads `last.nodes`, so carrying the breakdown on all 50 retained records grew every history file for no reader.
- Added a `from_result` test asserting BTreeMap node order, component values, and the duration/rows mapping, plus a test that older records shed node detail while metrics still render.
- README and `docs/current/ci-and-orchestration.md` document `duckle_node_last_duration_seconds` and `duckle_node_last_rows`.
- The `ctl.switch` claim in the body is corrected: a routed-around branch still executes on an empty relation and reports `rows: 0`; `None` is for a stage the run never reached.

SouravRoy-ETL:
All four verified on 203210e, including the one I cared about most: `nodes` is kept on the newest record only and the trim drops it from the rest, so the file stops growing without losing what `render_metrics` actually reads. CI is green on all three runners, and both new tests are real. The `from_result` one pins node order and the component value, which is exactly what a future refactor would have dropped silently.

Two nits, neither worth another round on its own:

Every run's history diff now also deletes the previous run's node block, so `runs/<id>.json` churns more per commit than it used to, in a file whose module doc sells small readable diffs. That is the right trade for the size fix. Just worth knowing it was made.

Storage is bounded now; the emitted series count is not. `render_metrics` still emits one series per node of the newest run per pipeline, and the README line about labels being bounded reads as though it covers the node dimension too. A sentence saying that dimension is bounded by pipeline size rather than by a cap would keep it honest.

Ready from my side.

